PR #2, opened 2020-08-06
cosa/20200805-meta-merge: design for merged meta.json

darkmuggle:
This describes the standard for merging `meta.json` for parallel and
distributed builds.

Signed-off-by: Ben Howard <ben.howard@redhat.com>

### DIFF
--- a/cosa/20200805-meta-merge.md
+++ b/cosa/20200805-meta-merge.md
@@ -1,0 +1,84 @@
+# Merging of build meta-data
+- Specification ID: cosa-20200805
+- Supersedes: None
+- Conflicts: None
+
+# Background
+
+One of the biggest blockers to performing parallel builds, distributed builds and/or re-starting failed builds is the ability to lock write access to `meta.json` in COSA.
+
+Under the following assumptions, COSA needs to support merging of `meta.json`:
+- Not all `cmd-buildextend-*` operations will happen on the same pod or Linux namespace
+- S3 is the authoritative store for builds. S3 is eventually consistent meaning that the version read from S3 may be replaced after reading from S3. Implementing distributed locking on `meta.json` based on S3 is problematic.
+- For RHCOS we need the ability to do partial builds and then back-fill artifacts (e.g. only build qemu, but not cloud) and later build cloud.
+- Developers want to restart failed builds
+- Consumers of the artifacts from COSA want to reduce the time to build and test
+
+## State of a build is a point-of-view statement
+
+To COSA a build is usually `ostree` + `qcow2`. The notion of a "failed" or "partial" build is a point of view. COSA does provision the build id, but COSA does not know what comprised the artifacts of the "build". Usually developers and consumers speak of a build as the base `ostree` and the _optional_ artifacts.
+
+In COSA a "build" only starts to exist when the `meta.json` is incepted after the `ostree` has been written. The build directory may exist, but the `meta.json` that is produced immediately is valid for any consumer. In most cases a "failed" build means that an artifact in the `buildextend-*` commands has failed to publish (such as AWS) not that the artifact itself failed to build; the artifact was indeed built.
+
+Further, a "failed build" is also used to mean that some part of a pipeline failed to complete. COSA itself does not have any concept of dependencies, much less the idea that a multiple COSA commands can be run. COSA only has implied dependencies that some artifacts requires others. In other words, COSA doesn't know when the "build" is complete until after something else considers it's complete.
+
+## Multiple `meta.json`
+
+Under the axiom that the state of a build is usually determined from the point-of-view of a pipeline and to support distributed and parallel builds, builds will be produced with multiple `meta.json` that will be merged together after the artifact is built.
+
+## `cosalib.meta` library
+
+`cmd-meta` currently uses the [`cosalib.meta`](https://github.com/coreos/coreos-assembler/blob/master/src/cosalib/meta.py) library for reading and writing `meta.json`. To support merging of `meta.json` files, the library will be updated to allow for merging in-memory updates with on-disk content.
+
+Only `meta.json` that is validated will be merged under the following rules:
+- source and destination meta must match the buildid, name and ostree commit
+- on-disk `meta.json` takes precedence over the in-memory meta
+- all versions must be validated before and after merging before commiting to disk.
+
+Prior to merging `cosalib.meta` will lock the localfile.
+
+## merge mode denoted in `meta.json`
+
+To support production of meta meta data, `cmd-build` will have a key to support merging meta-data in a delayed fashion.
+
+```
+{
+    	"coreos-assembler.delayed-meta-merge": true
+}
+```
+
+`cmd-build` will have the flag of `--delay-meta-merge` added to set the merge mode. The default is `false`.
+
+## Delayed Merging of meta-data
+
+When `coreos-assembler.delayed-meta-merge` is set to `true`, `cmd-buildextend.<ARTIFACT>` commands will create suffixed `meta.<ARTIFACT>.json` instead of writing directly to `meta.json`. A new command called `cmd-meta-merge` will be introduced that will read `meta.<ARTIFACT>.json` and will ONLY combine the images and the cloud-build specific sections into `meta.json`.
+
+Orchestration layers will be responsible for shuffling artifacts around, e.g. `cmd-upload-build`. `cmd-upload-build --meta <ARTIFACT>` will only upload the artifacts described in the meta.
+
+All `meta.<ARTIFACT>.json` will be removed when their contents are merged into `meta.json`. If the artifacts are produced across compute domains, orchestration layers are responsible for calling `cmd-upload-build` to stage the artifacts.
+
+`cmd-meta-merge` will be able to read `meta.<ARTIFACT>.json`. `cmd-meta-merge` will:
+- read in `meta.json` and validate it.
+- iterate over `meta.<ARTIFACT>.json` in the path or in S3.
+- Validation against the schema for both source and destination meta-data before and after merging.
+- the `meta.<ARTIFACT>.json` will be deleted
+
+## Merging logic
+
+By default, only keys not found in `meta.json` will be added from `meta.<ARTIFACT>.json`. To support rebuilds/restart, a force option will exist in `cmd-meta-merge` with a default of `false`.
+
+Keys that will be merged:
+- fields from the images section
+- known cloud names such as AWS, Azure, GPC, etc.
+
+## Default builds
+
+When `coreos-assembler.delayed-meta-merge` is false or absent, `cmd-buildextend-*` commands will will produce a `meta.<ARTIFACT>.json` and immediately merge into `meta.json` using the `cmd-meta-merge`. To support bash commands `finalize-artifact` will use `cmd-meta-meta` for `meta.json`.
+
+The immediate merging of metas requires:
+- time sync clocks
+- shared disk for `builds/`
+
+## Validation Requirement
+
+All merged `meta.json` must be validated against the schema.


darkmuggle:
This originally came out of the desire reflect the state of a build (being failed or partial) in the `meta.json`. The notion of build state comes the pipeline running in stages. And since COSA has no coherent idea of a hierarchy, dependencies or even that there are things called pipelines, baking in that a build is "failed" or "partial" seems out-of-place. Only the caller of the dozen or so COSA commands would know if the build was failed or partial. To COSA a build might be complete after the qcow2 or even the ostree repo has built. But to the pipeline a build might need the buildextend artifacts.

In short, my position is that only the pipelines would know that a build has started, failed, or is partial. COSA doesn't know what artifacts are wanted for that build. In fact, the idea of a build is usually taken as all the artifacts not what COSA itself considers a build. And since there was strong developer preference to make the code base hackable, COSA itself does not have the foundation that would provide the ability to know state.

To that end, my position is that:
- pipelines should handle the notion of "partial" or "failed"
- COSA should support emit multiple `meta.json` and their merging

In short, we are paying for not having a cogent design.

cosa/20200805-meta-merge.md
```diff
@@ -0,0 +1,103 @@
...
+- Conflicts: None
+
+# Background
+
+One of the biggest blockers to performing parallel, diustributed and/or re-starting failed builds is the ability to lock write access to `meta.json` in COSA. The 
+
+Under the following assmptions, COSA needs to support merging of `meta.json`:
+- Not all `cmd-buildextend-*` operations will happen on the same pod or Linux namespace
+- S3 is the authoritative store for builds. S3 is eventually consistent meaning that the version read from S3 may be replaced after reading from S3. Implementing distributed locking on `meta.json` based on S3 is problematic.
+- For RHCOS we the ability to do partial builds and then back-fill artifacts (e.g. only build qemu, but not cloud) and later build cloud.
```
ashcrow:
```suggestion
- For RHCOS we need the ability to do partial builds and then back-fill artifacts (e.g. only build qemu, but not cloud) and later build cloud.
```

cosa/20200805-meta-merge.md
```diff
@@ -0,0 +1,103 @@
...
+
+In COSA a "build" only starts to exist when the `meta.json` is incepted after the `ostree` has been written. The build directory may exists, but the `meta.json` that is produced immediately is valid for any consumers. In most cases a "failed" build means that an artifact in the `buildextend-*` commands has failed to publish (such as AWS) not that the artifact itself failed to build; the artifact was indeed built. 
+
+In most cases a "failed build" is used to mean that some part of a pipeline failed to complete. COSA itself does not have any concept of dependencies, much less the idea that a multiple COSA commands can be run. COSA only has implied dependencies that some artifacts requires others. In other words, COSA doesn't know when the "build" is complete until after something else considers its complete. 
+
+Denoting the state of a build in the `meta.json` will be owned by process/thing/person is orchestrating the build.
```
ashcrow:
```suggestion
Denoting the state of a build in the `meta.json` will be owned by process/thing/person that is orchestrating the build. 
```

cosa/20200805-meta-merge.md
```diff
@@ -0,0 +1,103 @@
...
+
+Denoting the state of a build in the `meta.json` will be owned by process/thing/person is orchestrating the build. 
+
+## Mutliple `meta.json`
+
+Under the axiom that the state of a build is usually from the point-of-view of a pipeline and to support distributed and parallel builds, builds will be produced with multiple `meta.json` that will be merged together.
+
+## `cosalib.meta` library
+
+`cmd-meta` currently uses the `cosalib.meta` library for reading and writing `meta.json`. To support merging of `meta.json` files, the library will be updated to allow for merging in-memory updates with on-disk.
```
ashcrow:
```suggestion
`cmd-meta` currently uses the `cosalib.meta` library for reading and writing `meta.json`. To support merging of `meta.json` files, the library will be updated to allow for merging in-memory updates with on-disk content.
```

ashcrow (approved):
Love it. I think this would be super helpful as we start moving the pipeline(s) into a distributed model.

darkmuggle:
WIP PR showing simple locking on `meta.json` for parallel mode filed against COSA. The lock implementation is safe for NFS and can be used across processes.

cosa/20200805-meta-merge.md
```diff
@@ -0,0 +1,103 @@
...
+}
+```
+
+`cmd-build` will have the flag of `--distributed` added to set the build mode.
+
+## Distributed builds
+
+When `cosa.buildmode` is set to `distributed`, `cmd-buildextend.<ARTIFACT>` commands will create suffixed `meta.json.<ARTIFACT>` instead of writing directly to `meta.json`. A new command called `cmd-meta-merge` will be introduced that will read `meta.json.<ARTIFACT>` and will ONLY combine the images and the cloud-build specific sections into `meta.json`. 
+
+Pipelines orchestrating the build will be responsible shuffling artifacts around.
```
LorbusChris:
```suggestion
Pipelines orchestrating the build will be responsible for shuffling artifacts around. 
```

darkmuggle:
Given that this design is blocking and there is one approval, I'm going to wait until next week and take silence as consent.  

@jlebon and @dustymabe can you please take a look? The "parallel" mode could make FCOS builds faster.

cosa/20200805-meta-merge.md
```diff
@@ -0,0 +1,103 @@
...
+
+## State of a build is a point-of-view statement
+
+To COSA a build is usually `ostree` + `qcow2`. The notion of a "failed" or "partial" build is a point of view. COSA does provision the build id, but COSA does not know what comprised the artifacts of the "build". Usually developers and consumers speak of a build as the base `ostree` and the _optional_ artifacts. 
+
+In COSA a "build" only starts to exist when the `meta.json` is incepted after the `ostree` has been written. The build directory may exist, but the `meta.json` that is produced immediately is valid for any consumer. In most cases a "failed" build means that an artifact in the `buildextend-*` commands has failed to publish (such as AWS) not that the artifact itself failed to build; the artifact was indeed built. 
+
+Further, a "failed build" is also used to mean that some part of a pipeline failed to complete. COSA itself does not have any concept of dependencies, much less the idea that a multiple COSA commands can be run. COSA only has implied dependencies that some artifacts requires others. In other words, COSA doesn't know when the "build" is complete until after something else considers its complete.
```
dustymabe:
```suggestion
Further, a "failed build" is also used to mean that some part of a pipeline failed to complete. COSA itself does not have any concept of dependencies, much less the idea that a multiple COSA commands can be run. COSA only has implied dependencies that some artifacts requires others. In other words, COSA doesn't know when the "build" is complete until after something else considers it's complete. 
```

Or maybe `considers it to be complete`.

cosa/20200805-meta-merge.md
```diff
@@ -0,0 +1,103 @@
...
+
+In COSA a "build" only starts to exist when the `meta.json` is incepted after the `ostree` has been written. The build directory may exist, but the `meta.json` that is produced immediately is valid for any consumer. In most cases a "failed" build means that an artifact in the `buildextend-*` commands has failed to publish (such as AWS) not that the artifact itself failed to build; the artifact was indeed built. 
+
+Further, a "failed build" is also used to mean that some part of a pipeline failed to complete. COSA itself does not have any concept of dependencies, much less the idea that a multiple COSA commands can be run. COSA only has implied dependencies that some artifacts requires others. In other words, COSA doesn't know when the "build" is complete until after something else considers its complete. 
+
+Denoting the state of a build in the `meta.json` will be owned by process/thing/person that is orchestrating the build. 
+
+## Mutliple `meta.json`
```
dustymabe:
```suggestion
## Multiple `meta.json`
```

cosa/20200805-meta-merge.md
```diff
@@ -0,0 +1,103 @@
...
+
+This mode assumes that the build happens within the same kernel namespace. 
+
+## Validation Requirement
+
+All merged `meta.json` must be validated against the schema.
+
+## Publication
+
+Orchestrating pipelines are responsible for only publishing builds to the canonical locations.
```
dustymabe: Can you go in a little more detail here? I think I know what you mean but I'm not sure I follow 100%.

cosa/20200805-meta-merge.md
```diff
@@ -0,0 +1,103 @@
...
+Only `meta.json` that is validated will be merged under the following rules:
+- the on-disk version must match in core details for name and buildid for in memory model of the meta
+- on-disk meta.json takes precedence over the in-memory meta
+- all versions must be validated before and after merging before commiting to disk.
+
+Prior to merging `cosalib.meta` will lock the localfile.
+
+## build mode denoted in `meta.json`
+
+Two new build "modes" will be added. `cmd-build` will have build modes written in the `meta.json` of `parallel` (default) or distributed.
```
dustymabe: bikeshed: does `parallel` really mean anything here? I only really know the FCOS pipeline but I'll still ask the question: Today are our steps really done in parallel?

`distributed` makes sense, but I would almost say that executing a distributed build is the way to do parts of the pipeline in parallel.

cosa/20200805-meta-merge.md
```diff
@@ -0,0 +1,103 @@
+# Merging of build meta-data
+- Specification ID: cosa-20200804
+- Supersedes: None
+- Conflicts: None
+
+# Background
+
+One of the biggest blockers to performing parallel, diustributed and/or re-starting failed builds is the ability to lock write access to `meta.json` in COSA.
```
dustymabe:
```suggestion
One of the biggest blockers to performing parallel, distributed and/or re-starting failed builds is the ability to lock write access to `meta.json` in COSA.
```

cosa/20200805-meta-merge.md
```diff
@@ -0,0 +1,103 @@
...
+
+`cmd-build` will have the flag of `--distributed` added to set the build mode.
+
+## Distributed builds
+
+When `cosa.buildmode` is set to `distributed`, `cmd-buildextend.<ARTIFACT>` commands will create suffixed `meta.json.<ARTIFACT>` instead of writing directly to `meta.json`. A new command called `cmd-meta-merge` will be introduced that will read `meta.json.<ARTIFACT>` and will ONLY combine the images and the cloud-build specific sections into `meta.json`. 
+
+Pipelines orchestrating the build will be responsible for shuffling artifacts around. 
+
+All `meta.json.<ARTIFACT>` files will be archived in S3 alongside the `meta.json`.
```
dustymabe: What do you think about putting them in a sub-directory just to keep the toplevel dir cleaner?

cosa/20200805-meta-merge.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Distributed builds
+
+When `cosa.buildmode` is set to `distributed`, `cmd-buildextend.<ARTIFACT>` commands will create suffixed `meta.json.<ARTIFACT>` instead of writing directly to `meta.json`. A new command called `cmd-meta-merge` will be introduced that will read `meta.json.<ARTIFACT>` and will ONLY combine the images and the cloud-build specific sections into `meta.json`. 
+
+Pipelines orchestrating the build will be responsible for shuffling artifacts around. 
+
+All `meta.json.<ARTIFACT>` files will be archived in S3 alongside the `meta.json`. 
+
+`cmd-meta-merge` will be able `meta.json.<ARTIFACT>` located directly in S3. `cmd-meta-merge` will:
```
dustymabe: will be able.... missing word?

cosa/20200805-meta-merge.md
```diff
@@ -0,0 +1,103 @@
...
+{ 
+    "cosa.buildmode": "<parallel|distributed>"
+}
+```
+
+`cmd-build` will have the flag of `--distributed` added to set the build mode.
+
+## Distributed builds
+
+When `cosa.buildmode` is set to `distributed`, `cmd-buildextend.<ARTIFACT>` commands will create suffixed `meta.json.<ARTIFACT>` instead of writing directly to `meta.json`. A new command called `cmd-meta-merge` will be introduced that will read `meta.json.<ARTIFACT>` and will ONLY combine the images and the cloud-build specific sections into `meta.json`.
```
dustymabe: Is `cmd-meta-merge` a new pipeline stage essentially that will run at the end?

cosa/20200805-meta-merge.md
```diff
@@ -0,0 +1,103 @@
...
+- Supersedes: None
+- Conflicts: None
+
+# Background
+
+One of the biggest blockers to performing parallel, diustributed and/or re-starting failed builds is the ability to lock write access to `meta.json` in COSA.
+
+Under the following assmptions, COSA needs to support merging of `meta.json`:
+- Not all `cmd-buildextend-*` operations will happen on the same pod or Linux namespace
+- S3 is the authoritative store for builds. S3 is eventually consistent meaning that the version read from S3 may be replaced after reading from S3. Implementing distributed locking on `meta.json` based on S3 is problematic.
```
dustymabe, line 12: So how do we deal with this problem specifically. In the new flow do we:

- create meta.json upon first "build" (ostree + qcow2) upload to s3 (first time it exists in s3)
- create `meta.json.<ARTIFACT>` files
- `cmd-meta-merge` (pull `meta.json` from s3 and then merge in new info and then re-upload)

So the 3rd step would only be able to pull `meta.json` if it exists and there were no other updates to it so the eventual consistency problem isn't an issue?

cosa/20200805-meta-merge.md
miabbott:
```suggestion
One of the biggest blockers to performing parallel builds, distributed builds and/or re-starting failed builds is the ability to lock write access to `meta.json` in COSA.
```

cosa/20200805-meta-merge.md
miabbott:
```suggestion
## Multiple `meta.json`
```

cosa/20200805-meta-merge.md
```diff
@@ -0,0 +1,103 @@
...
+
+Denoting the state of a build in the `meta.json` will be owned by process/thing/person that is orchestrating the build. 
+
+## Mutliple `meta.json`
+
+Under the axiom that the state of a build is usually determined from the point-of-view of a pipeline and to support distributed and parallel builds, builds will be produced with multiple `meta.json` that will be merged together.
+
+## `cosalib.meta` library
+
+`cmd-meta` currently uses the `cosalib.meta` library for reading and writing `meta.json`. To support merging of `meta.json` files, the library will be updated to allow for merging in-memory updates with on-disk content.
```
miabbott:
```suggestion
`cmd-meta` currently uses the [`cosalib.meta`](https://github.com/coreos/coreos-assembler/blob/master/src/cosalib/meta.py) library for reading and writing `meta.json`. To support merging of `meta.json` files, the library will be updated to allow for merging in-memory updates with on-disk content.
```

cosa/20200805-meta-merge.md
```diff
@@ -0,0 +1,103 @@
...
+
+When `cosa.buildmode` is set to `distributed`, `cmd-buildextend.<ARTIFACT>` commands will create suffixed `meta.json.<ARTIFACT>` instead of writing directly to `meta.json`. A new command called `cmd-meta-merge` will be introduced that will read `meta.json.<ARTIFACT>` and will ONLY combine the images and the cloud-build specific sections into `meta.json`. 
+
+Pipelines orchestrating the build will be responsible for shuffling artifacts around. 
+
+All `meta.json.<ARTIFACT>` files will be archived in S3 alongside the `meta.json`. 
+
+`cmd-meta-merge` will be able `meta.json.<ARTIFACT>` located directly in S3. `cmd-meta-merge` will:
+- read in `meta.json` and validate it.
+- iterate over the remaining `meta.json.<ARTIFACT>`. Validation against the schema will happen during the read and after merging.
```
miabbott: Will `cmd-meta-merge` also push the result to S3?

cosa/20200805-meta-merge.md
miabbott:
```suggestion
`cmd-meta-merge` will be able to read `meta.json.<ARTIFACT>` located directly in S3. `cmd-meta-merge` will:
```

cosa/20200805-meta-merge.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Local builds
+
+When `cosa.buildmode` is set to "parallel" `cmd-buildextend-*` the merging of `meta.json` will happen in real time using the `cosalib.meta` Python code to prevent racing conditions. To support bash commands `finalize-artifact` will be replaced with `cmd-meta-merge`.
```
miabbott:
```suggestion
When `cosa.buildmode` is set to "parallel", the `cmd-buildextend-*` commands will manage the merging of `meta.json` in real time using the `cosalib.meta` Python code to prevent racing conditions. To support bash commands `finalize-artifact` will be replaced with `cmd-meta-merge`.
```

Not sure this was the intent or not

cosa/20200805-meta-merge.md
```diff
@@ -0,0 +1,103 @@
...
+```
+
+This meta-data section is indended for use by publishers. COSA will:
+- ignore the section entirely
+- the schema will have it added as an optional, albeit ignored field.
+
+And example might be:
+```
+{
+    "publiser": {
```
miabbott:
```suggestion
    "publisher": {
```

cosa/20200805-meta-merge.md
```diff
@@ -0,0 +1,103 @@
...
+
+Pipelines orchestrating the build will be responsible for shuffling artifacts around. 
+
+All `meta.json.<ARTIFACT>` files will be archived in S3 alongside the `meta.json`. 
+
+`cmd-meta-merge` will be able `meta.json.<ARTIFACT>` located directly in S3. `cmd-meta-merge` will:
+- read in `meta.json` and validate it.
+- iterate over the remaining `meta.json.<ARTIFACT>`. Validation against the schema will happen during the read and after merging.
+
+## Local builds
```
miabbott: Maybe change the title of this section to `Parallel builds` to match the `buildmode` option

darkmuggle: Good call. Updated.

cosa/20200805-meta-merge.md
```diff
@@ -0,0 +1,103 @@
+# Merging of build meta-data
+- Specification ID: cosa-20200804
```
dustymabe:
```suggestion
- Specification ID: cosa-20200805
```

dustymabe (commented):
Thanks for putting this together! I've got some questions just to try to clear some things up in my head about how this all fits together.

cosa/20200805-meta-merge.md
```diff
@@ -0,0 +1,103 @@
...
+
+# Background
+
+One of the biggest blockers to performing parallel builds, distributed builds and/or re-starting failed builds is the ability to lock write access to `meta.json` in COSA.
+
+Under the following assmptions, COSA needs to support merging of `meta.json`:
+- Not all `cmd-buildextend-*` operations will happen on the same pod or Linux namespace
+- S3 is the authoritative store for builds. S3 is eventually consistent meaning that the version read from S3 may be replaced after reading from S3. Implementing distributed locking on `meta.json` based on S3 is problematic.
+- For RHCOS we need the ability to do partial builds and then back-fill artifacts (e.g. only build qemu, but not cloud) and later build cloud.
+- Developers want to restart failed builds
```
arithx, line 14: Probably worth distinguishing that we want to be able to restart failed build pipeline steps (e.x.: re-run $platform kola)

cosa/20200805-meta-merge.md
```diff
@@ -0,0 +1,103 @@
...
+- Consumers of the artifacts from COSA want to reduce the time to build and test
+
+## State of a build is a point-of-view statement
+
+To COSA a build is usually `ostree` + `qcow2`. The notion of a "failed" or "partial" build is a point of view. COSA does provision the build id, but COSA does not know what comprised the artifacts of the "build". Usually developers and consumers speak of a build as the base `ostree` and the _optional_ artifacts. 
+
+In COSA a "build" only starts to exist when the `meta.json` is incepted after the `ostree` has been written. The build directory may exist, but the `meta.json` that is produced immediately is valid for any consumer. In most cases a "failed" build means that an artifact in the `buildextend-*` commands has failed to publish (such as AWS) not that the artifact itself failed to build; the artifact was indeed built.
```
arithx: Given we've established in the previous paragraph that the definition of "failed" or "partial" builds depends on the context / point of view we should probably include context information when mentioning `a "failed" build`.

cosa/20200805-meta-merge.md
```diff
@@ -0,0 +1,103 @@
...
+
+In COSA a "build" only starts to exist when the `meta.json` is incepted after the `ostree` has been written. The build directory may exist, but the `meta.json` that is produced immediately is valid for any consumer. In most cases a "failed" build means that an artifact in the `buildextend-*` commands has failed to publish (such as AWS) not that the artifact itself failed to build; the artifact was indeed built. 
+
+Further, a "failed build" is also used to mean that some part of a pipeline failed to complete. COSA itself does not have any concept of dependencies, much less the idea that a multiple COSA commands can be run. COSA only has implied dependencies that some artifacts requires others. In other words, COSA doesn't know when the "build" is complete until after something else considers its complete. 
+
+Denoting the state of a build in the `meta.json` will be owned by process/thing/person that is orchestrating the build.
```
arithx:
```suggestion
Denoting the state of a build in the `meta.json` will be owned by the orchestration layer. 
```

cosa/20200805-meta-merge.md
```diff
@@ -0,0 +1,103 @@
...
+
+Further, a "failed build" is also used to mean that some part of a pipeline failed to complete. COSA itself does not have any concept of dependencies, much less the idea that a multiple COSA commands can be run. COSA only has implied dependencies that some artifacts requires others. In other words, COSA doesn't know when the "build" is complete until after something else considers its complete. 
+
+Denoting the state of a build in the `meta.json` will be owned by process/thing/person that is orchestrating the build. 
+
+## Multiple `meta.json`
+
+Under the axiom that the state of a build is usually determined from the point-of-view of a pipeline and to support distributed and parallel builds, builds will be produced with multiple `meta.json` that will be merged together.
```
arithx:
```suggestion
Under the axiom that the state of a build is usually determined from the point-of-view of a pipeline and to support distributed and parallel builds, builds will be produced with multiple `meta.json` stubs that will be merged together.
```

cosa/20200805-meta-merge.md
```diff
@@ -0,0 +1,103 @@
...
+
+## `cosalib.meta` library
+
+`cmd-meta` currently uses the [`cosalib.meta`](https://github.com/coreos/coreos-assembler/blob/master/src/cosalib/meta.py) library for reading and writing `meta.json`. To support merging of `meta.json` files, the library will be updated to allow for merging in-memory updates with on-disk content.
+
+Only `meta.json` that is validated will be merged under the following rules:
+- the on-disk version must match in core details for name and buildid for in memory model of the meta
+- on-disk meta.json takes precedence over the in-memory meta
```
arithx: Is this implying for non-core details? Otherwise how would we end up in skew if the on-disk version must match the in-memory model?

cosa/20200805-meta-merge.md
```diff
@@ -0,0 +1,103 @@
...
+- all versions must be validated before and after merging before commiting to disk.
+
+Prior to merging `cosalib.meta` will lock the localfile.
+
+## build mode denoted in `meta.json`
+
+Two new build "modes" will be added. `cmd-build` will have build modes written in the `meta.json` of `parallel` (default) or distributed.
```
arithx: Hmm, I wonder if there wouldn't be some value in always producing/merging stubs and instead having a command (or workflow step) that fetches the stubs when ran in a distributed fashion?

cosa/20200805-meta-merge.md
darkmuggle: The difference between `parallel` and `distributed` is to tell COSA when the merge should take place. In `parallel` COSA would assume a common file-system (NFS?) were `distributed` assumes neither time nor place is common; the same ostree being extended in time in different logical domains (think different OCP clusters, pods, and even clouds).

The other reason for having distributed mode is that we could even re-create artifacts later from an ostree.  For example, build `oscontainers` and if we need boot images, back-fill them; artifacts could conceivably be incepted after publication.

cosa/20200805-meta-merge.md
darkmuggle: Honestly, I don't really have a strong opinion either way, but I do like that it's clean. The only reason I proposed archiving `meta.json.<ARTIFACT>` was to provide for the ability to start testing before things are finished. If we have `meta.json.<CLOUD>` appear while other artifacts are being produced, we could conceivably start end-to-end testing earlier. 

But sub-dir seems sensible to me.

cosa/20200805-meta-merge.md
darkmuggle: Yeah, it would be a new stage called by the orchestrator the build. It would only need to be called by `distributed` builds.

cosa/20200805-meta-merge.md
darkmuggle: Correct.  Since the `meta.json` is only ever written twice, the race for eventual consistency is mitigated. I envision that in distributed mode the orchestrator would despot the artifact into S3 itself with `meta.json.<ARTIFACT>`.  Since `cmd-upload-build` won't overwrite the files that exist. So the flow would be:
- create base build and run `cmd-upload-build`
- distributed builds for $ARTIFACT and call `cmd-upload-build` at the end of each
- sync `meta.json.<ARTIFACT>`  and run `cmd-meta-merge` and upload final `meta.json`

cosa/20200805-meta-merge.md
```diff
@@ -0,0 +1,103 @@
+# Merging of build meta-data
+- Specification ID: cosa-20200805
+- Supersedes: None
+- Conflicts: None
+
+# Background
+
+One of the biggest blockers to performing parallel builds, distributed builds and/or re-starting failed builds is the ability to lock write access to `meta.json` in COSA.
+
+Under the following assmptions, COSA needs to support merging of `meta.json`:
```
jlebon:
```suggestion
Under the following assumptions, COSA needs to support merging of `meta.json`:
```

cosa/20200805-meta-merge.md
jlebon: > create base build and run `cmd-upload-build`

How about:

> create base build and run qemu tests and run `cmd-upload-build`

? :) 

> - sync `meta.json.<ARTIFACT>` and run `cmd-meta-merge` and upload final `meta.json`

Should we have a step after that for deleting the `meta.json.<ARTIFACT>` files which were merged?

cosa/20200805-meta-merge.md
```diff
@@ -0,0 +1,103 @@
...
+
+Further, a "failed build" is also used to mean that some part of a pipeline failed to complete. COSA itself does not have any concept of dependencies, much less the idea that a multiple COSA commands can be run. COSA only has implied dependencies that some artifacts requires others. In other words, COSA doesn't know when the "build" is complete until after something else considers it's complete. 
+
+Denoting the state of a build in the `meta.json` will be owned by process/thing/person that is orchestrating the build. 
+
+## Multiple `meta.json`
+
+Under the axiom that the state of a build is usually determined from the point-of-view of a pipeline and to support distributed and parallel builds, builds will be produced with multiple `meta.json` that will be merged together.
```
jlebon: I wonder if the semantics we should be aiming for here is "one primary `meta.json` and multiple `meta.json.<FOO>` JSON fragments; merged together in the way of `jq -s add`". So e.g. the `meta.json.openstack` would *just* be:

```json
{
	"images": {
		"openstack": {
			...
		}
	}
}
```

That way, there is no room for misinterpretation or for multiple sources of truth, etc... and it makes the merging step dead simple.

cosa/20200805-meta-merge.md
```diff
@@ -0,0 +1,103 @@
...
+
+## `cosalib.meta` library
+
+`cmd-meta` currently uses the [`cosalib.meta`](https://github.com/coreos/coreos-assembler/blob/master/src/cosalib/meta.py) library for reading and writing `meta.json`. To support merging of `meta.json` files, the library will be updated to allow for merging in-memory updates with on-disk content.
+
+Only `meta.json` that is validated will be merged under the following rules:
+- the on-disk version must match in core details for name and buildid for in memory model of the meta
+- on-disk meta.json takes precedence over the in-memory meta
```
jlebon:
```suggestion
- on-disk `meta.json` takes precedence over the in-memory meta
```

For consistency with the rest of the doc

cosa/20200805-meta-merge.md
```diff
@@ -0,0 +1,103 @@
...
+
+Prior to merging `cosalib.meta` will lock the localfile.
+
+## build mode denoted in `meta.json`
+
+Two new build "modes" will be added. `cmd-build` will have build modes written in the `meta.json` of `parallel` (default) or distributed.
+
+```
+{ 
+    "cosa.buildmode": "<parallel|distributed>"
```
jlebon: To be consistent with other cosa-related keys, I'd make this `coreos-assembler.buildmode`.

cosa/20200805-meta-merge.md
jlebon: Hmm, I see what @dustymabe means. `parallel` could seem misleading in this context because it's still up to you to orchestrate cosa to run in parallel and cosa itself doesn't really need special knowledge/behaviour in that mode (once we have the locking work in place, but that'll just always be on).

Conversely, in `distributed` mode, cosa is just being helpful about how to spit out JSON files, but it's not doing any orchestration there too.

My bikeshed: we make it a boolean and rename it to something a bit more descriptive, e.g.:

```json
{
	"coreos-assembler.delayed-meta-merge": true
}
```

cosa/20200805-meta-merge.md
dustymabe: Sounds good. It might be worth adding just a little more detail in here to spell that out. It's subtle, but important.

cosa/20200805-meta-merge.md
jlebon: I made a comment higher up about deleting the `meta.json.<ARTIFACT>` files after merging. I don't like the idea of having duplicate data in multiple metadata files.

I like your idea about early testing. It seems like whatever tests we want to run before merge could fetch both `meta.json` and try to fetch `meta.json.<ARTIFACT>` but allow it not to exist?

dustymabe: > Honestly, I don't really have a strong opinion either way, but I do like that it's clean. 

👍

> The only reason I proposed archiving `meta.json.<ARTIFACT>` was to provide for the ability to start testing before things are finished. If we have `meta.json.<CLOUD>` appear while other artifacts are being produced, we could conceivably start end-to-end testing earlier.

I don't think what I'm proposing would limit that. We could use the same subdir for all "distributed" `meta.json.<ARTIFACT>` files. Kind of like we have a [manifests dir](https://github.com/coreos/fedora-coreos-config/tree/testing-devel/manifests) in our fcos configs, and then a toplevel manifest that includes them. It's just a logical grouping of like things, but keeps the number of files at the toplevel down so it doesn't overwhelm someone when they first look at it.

cosa/20200805-meta-merge.md
```diff
@@ -0,0 +1,103 @@
...
+
+`cmd-meta-merge` will be able to read `meta.json.<ARTIFACT>` located directly in S3. `cmd-meta-merge` will:
+- read in `meta.json` and validate it.
+- iterate over the remaining `meta.json.<ARTIFACT>`. Validation against the schema will happen during the read and after merging.
+
+## Parallel builds
+
+When `cosa.buildmode` is set to "parallel", the `cmd-buildextend-*` commands will manage the merging of `meta.json` in real time using the `cosalib.meta` Python code to prevent racing conditions. To support bash commands `finalize-artifact` will be replaced with `cmd-meta-merge`.
+
+This mode assumes that the build happens within the same kernel namespace.
```
jlebon: I think we can loosen this to just assuming that the builds share a common `builds/` directory over which they can use locking.

cosa/20200805-meta-merge.md
dustymabe: proposal: rename `parallel` in your proposal to `default`

my understanding is that `parallel` is just what we have today plus some locking around the `meta.json` writes, which is probably useful on its own (as mentioned in https://github.com/coreos/coreos-assembler/pull/1644#issuecomment-672336739) and is what would be used in local development and also in our pipelines at least until we started to take advantage of `distributed` mode.

cosa/20200805-meta-merge.md
```diff
@@ -0,0 +1,103 @@
...
+
+And example might be:
+```
+{
+    "publisher": {
+        "state": "done",
+        "developer-build": False
+    }
+}
+```
```
jlebon: Hmm, do you have more context on this one? It sounds like pipeline implementation details leaking into the schema. Can we drop this part of the proposal until it's actually needed?

cosa/20200805-meta-merge.md
darkmuggle: Next commit has this resolved.

cosa/20200805-meta-merge.md
darkmuggle: Struck the paragraph given the feedback to remove the publisher setting

cosa/20200805-meta-merge.md
darkmuggle: Interesting idea... I will explore that, but to use a stub I need to rework the design. I'm trying to think through the implications of what that would mean. The obvious implications get a little funny around the "when" since the default build mode would call it. Are you thinking that there would be `cmd-meta-merge --final` for the `coreos-assembler.delay-meta-merge`?

cosa/20200805-meta-merge.md
darkmuggle: Updated. Since this specific to COSA, I imagine that the FCOS and RHCOS pipelines will handle publication differently.

cosa/20200805-meta-merge.md
darkmuggle: The idea here was that you can't merge FCOS into RHCOS; that why I called out `name` and `buildid`. I wanted to prevent merging 20200101 into 202501010.

cosa/20200805-meta-merge.md
darkmuggle: I've updated the spec for a `stub.<artifact>.json`. But to be honest, I do not like introducing another data-model. The advantage of having complete `meta.json` is that it follows the schema and that means things like `kola` understand it. The other reason I like having the complete `meta.json.<ARTIFACT>` is that it gives the ability to ensure that the builds are the same. 

After thinking this through the only way I could get on board is to introduce another schema to ensure that the artifacts are from the same build -- we can compare the name, build id and ostree to ensure that its the same.

Relying on `jq -s add` may be simple for the bash case, but that won't work with the `cosalib.meta` case. After looking at the code, this is actually harder to implement for the Pythonic code. 

> That way, there is no room for misinterpretation or for multiple sources of truth, etc... and it makes the merging step dead simple.

Merging is dead simple for bash, not so much for Python. If, per the other comments, we are not publishing the artifacts JSON, then what will mis-interpret it? We could use a different name, e.g. `stub.<ARTIFACT>.json` to reflect its incomplete. Today the tooling only understands `builds.json` and `meta.json`. So the only case that I can think of is the developer case. 

I've updated the proposal to reflect this idea, but I'm not a fan.

cosa/20200805-meta-merge.md
darkmuggle: 👍

cosa/20200805-meta-merge.md
darkmuggle: dropped

cosa/20200805-meta-merge.md
jlebon: Ack, thanks for considering. I think following the schema is totally fine if it simplifies the implementation. If we're agreed on deleting the files, as you said there's nothing else that should consume it anyway. So going back to non-fragments SGTM!

darkmuggle: What do you think of using a naming convention of `artifacts.<ARTIFACT>.json`  (e.g. `artifacts.aws.json`)?

jlebon: I like that it ends with `.json` (nice for editors/highlighters). Another suggestion which is a slight tweak on your original proposal is `meta.<ARTIFACT>.json` so it stays close to `meta.json` in listings. WFM either way though.

jlebon (approved):
To make it official, LGTM! (Regardless of the bikeshed in https://github.com/coreos/enhancements/pull/2#discussion_r470096813).

darkmuggle:
@ashcrow @jlebon can I get a re-ack? 

I thought I had pushed the stub revert after https://github.com/coreos/enhancements/pull/2#discussion_r469562475

This is the final push before I'd like to start doing the work.

jlebon (approved):

ashcrow (approved):

dustymabe:
LGTM - thanks for working on this @darkmuggle

PR #3, opened 2020-11-03
os: add initial full-disk RAID proposal

bgilbert:
For https://github.com/coreos/fedora-coreos-tracker/issues/581.

### DIFF
--- a/os/20201103-full-disk-raid.md
+++ b/os/20201103-full-disk-raid.md
@@ -1,0 +1,382 @@
+# Full disk RAID support
+
+## Theory of operation
+
+Users can write FCC sugar that enables redundant boot and/or an encrypted root volume.  There are three cases:
+
+1. The sugar only specifies encrypted root.  FCCT desugars the section into a LUKS volume within the existing root partition, and a root filesystem layered on top.
+2. The sugar only specifies redundant boot.  FCCT desugars the section into an Ignition config which completely recreates the partition structure on each specified disk.  The config creates partitions, RAID volumes, and filesystems for the root, boot, and ESP partitions, and partitions for each BIOS-BOOT replica.
+3. The sugar specifies both.  FCCT desugars similarly to the second case, with the addition of a LUKS volume.
+
+If redundant boot is specified, Dracut glue detects the corresponding config stanzas, saves the corresponding on-disk data to RAM before the Ignition disks phase, and restores it after disks phase completes.  The glue also replicates the boot sector when replicating BIOS-BOOT.
+
+Each piece of the replication logic (root partition, boot partition, ESP, and BIOS-BOOT + MBR) functions independently.  An advanced user can create nonstandard partition configurations (for example, RAID-1 boot with RAID-5 root) by manually configuring partitions in their FCC.
+
+Recovery from a disk failure is out of scope for this enhancement.  We will recommend that users recover from a hardware failure by reprovisioning the machine.
+
+## FCC sugar
+
+```yaml
+variant: fcos
+version: 1.3.0
+boot_device:
+  # Allows choosing from multiple disk layout templates hardcoded in FCCT.
+  # We might use this to support CPU architectures with different partition
+  # requirements (e.g. ppc64le) or to change the disk layout in future OS
+  # releases.
+  # Optional, defaults to x86_64
+  layout: x86_64
+  # If specified, mirror every default partition
+  mirror:
+    devices:
+      # Two or more devices are required
+      - /dev/vda
+      - /dev/vdb
+  # If specified, encrypt the root partition
+  luks:
+    # The contents of the existing clevis struct, without the `custom`
+    # field (which is too advanced for sugar).  We don't need to support
+    # static key files because they don't work for root.
+    tang:
+      - url: https://example.com/tang.tang.tang
+        thumbprint: x
+    tpm2: true
+    threshold: 2
+```
+
+The `boot_device` struct is at the top level, rather than under `storage`, because CT sugar has historically been placed at the top level and because Go struct embedding semantics make it harder to put it under `storage`.
+
+The sugar does not support advanced features such as RAID spares, RAID levels other than 1, or filesystem/mount options.  Users can manually specify individual partitions, RAIDs, and filesystems if desired.
+
+## Rendered Ignition config
+
+### LUKS only
+
+```json
+{
+  "ignition": {
+    "version": "3.2.0"
+  },
+  "storage": {
+    "filesystems": [
+      {
+        "device": "/dev/disk/by-id/dm-name-root",
+        "format": "xfs",
+        "label": "root",
+        "wipeFilesystem": true
+      }
+    ],
+    "luks": [
+      {
+        "clevis": {
+          "tpm2": true
+        },
+        "device": "/dev/disk/by-partlabel/root",
+        "label": "luks-root",
+        "name": "root",
+        "wipeVolume": true
+      }
+    ]
+  }
+}
+```
+
+### RAID only
+
+```json
+{
+  "ignition": {
+    "version": "3.2.0"
+  },
+  "storage": {
+    "disks": [
+      {
+        "device": "/dev/vda",
+        "partitions": [
+          {
+            "label": "bios-1",
+            "sizeMiB": 1,
+            "typeGuid": "21686148-6449-6E6F-744E-656564454649"
+          },
+          {
+            "label": "esp-1",
+            "sizeMiB": 127,
+            "typeGuid": "C12A7328-F81F-11D2-BA4B-00A0C93EC93B"
+          },
+          {
+            "label": "boot-1",
+            "sizeMiB": 384
+          },
+          {
+            "label": "root-1"
+          }
+        ],
+        "wipeTable": true
+      },
+      {
+        "device": "/dev/vdb",
+        "partitions": [
+          {
+            "label": "bios-2",
+            "sizeMiB": 1,
+            "typeGuid": "21686148-6449-6E6F-744E-656564454649"
+          },
+          {
+            "label": "esp-2",
+            "sizeMiB": 127,
+            "typeGuid": "C12A7328-F81F-11D2-BA4B-00A0C93EC93B"
+          },
+          {
+            "label": "boot-2",
+            "sizeMiB": 384
+          },
+          {
+            "label": "root-2"
+          }
+        ],
+        "wipeTable": true
+      }
+    ],
+    "filesystems": [
+      {
+        "device": "/dev/md/md-boot",
+        "format": "ext4",
+        "label": "boot",
+        "wipeFilesystem": true
+      },
+      {
+        "device": "/dev/md/md-esp",
+        "format": "vfat",
+        "label": "EFI-SYSTEM",
+        "wipeFilesystem": true
+      },
+      {
+        "device": "/dev/md/md-root",
+        "format": "xfs",
+        "label": "root",
+        "wipeFilesystem": true
+      }
+    ],
+    "raid": [
+      {
+        "devices": [
+          "/dev/disk/by-partlabel/boot-1",
+          "/dev/disk/by-partlabel/boot-2"
+        ],
+        "level": "raid1",
+        "name": "md-boot",
+        "options": [
+          "--metadata=1.0"
+        ]
+      },
+      {
+        "devices": [
+          "/dev/disk/by-partlabel/esp-1",
+          "/dev/disk/by-partlabel/esp-2"
+        ],
+        "level": "raid1",
+        "name": "md-esp",
+        "options": [
+          "--metadata=1.0"
+        ]
+      },
+      {
+        "devices": [
+          "/dev/disk/by-partlabel/root-1",
+          "/dev/disk/by-partlabel/root-2"
+        ],
+        "level": "raid1",
+        "name": "md-root"
+      }
+    ]
+  }
+}
+```
+
+FCCT assigns a unique serial number to each disk and appends it to the label assigned to each disk partition.  This ensures each partition can be referenced by a unique path in `/dev/disk/by-partlabel`.
+
+When desugaring, FCCT should be careful to allow explicit settings elsewhere in the config to override settings originating from sugar.  This could be used e.g. to specify filesystem formatting options.
+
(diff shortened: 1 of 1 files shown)


os/20201103-full-disk-raid.md
```diff
@@ -0,0 +1,208 @@
...
+            "label": "root-2",
+            "sizeMiB": 8192
+          }
+        ],
+        "wipeTable": true
+      }
+    ],
+    "filesystems": [
+      {
+        "device": "/dev/disk/by-partlabel/esp-1",
```
Prashanth684: by-partlabel will not work on s390x : https://github.com/coreos/ignition-dracut/pull/100#issuecomment-529948403

The GPT partitioning scheme uses little endian by default whereas s390x is big endian causing the by-partlabel string to be interpreted incorrectly.

os/20201103-full-disk-raid.md
```diff
@@ -0,0 +1,208 @@
...
+
+For ppc64le, we'll create independent replicas of the PReP partition, similar to the BIOS boot partition.  We'll need to confirm with multi-arch folks that this will work.
+
+## Ignition changes
+
+The Dracut glue can detect the requisite Ignition config stanzas using `jq`, but that's brittle and we're already doing a lot of it in the initrd.  Ideally Ignition will provide command-line arguments for executing relevant (hard-coded) queries against the cached Ignition config.
+
+## Architecture support
+
+We'll need to get feedback from the multi-arch team for aarch64 and ppc64le.  ppc64le will need an alternate `layout` definition that includes the PReP partition.
```
Prashanth684: i think this would work. The only thing I remember being special about the Prep partition was it needs to be the first partition to work correctly on some systems such as powerVM.

os/20201103-full-disk-raid.md
```diff
@@ -0,0 +1,208 @@
...
+
+For s390x, Ignition doesn't yet support DASD partitioning, and we'd need provisions for rerunning zipl from the initramfs, if RAID support is even possible at all.  Each DASD is connected via a particular bus identifier, which might need to be encoded into each respective copy of the bootloader.  We may need to declare full-disk RAID out of scope for s390x, at least for 4.7.
```
Prashanth684: while this is true, there are s390x systems which use FCP disks for which this could be supported.

os/20201103-full-disk-raid.md
bgilbert: Okay, we can have FCCT set partition UUIDs and use `/dev/disk/by-partuuid` instead.

In the long run, is the s390x partlabel issue being fixed?

os/20201103-full-disk-raid.md
bgilbert: Sure, we could make that distinction.  But we'd still need to rerun zipl on FCP, right?  We're copying the boot filesystem as a tree, not an image, so the kernel/initrd blocklists will change.

os/20201103-full-disk-raid.md
```diff
@@ -0,0 +1,208 @@
...
+The BIOS Boot partition is also not updated at runtime, so RAID is not necessary here either.  As a result, when updating the bootloader, bootupd will need to find each BIOS Boot type GUID and update both the partition and the corresponding boot sector.  Again this assumes that the entire system belongs to CoreOS.
+
+For simplicity, we're copying the BIOS Boot partition and boot sector bit-for-bit, rather than rerunning `grub-install` from the initrd or hand-modifying the boot sector.  Since the boot sector hardcodes the offset of the BIOS Boot partition, the latter must be recreated at the same offset as the original partition.  To avoid hardcoding an awkward constant in FCCT, we'll move the BIOS boot partition from its current offset of 512 MiB to the beginning of the disk (offset 1 MiB) in new images.  The Dracut glue will detect relocation of the BIOS boot partition and fail the boot.
+
+There's no BIOS boot partition in 4Kn images.  We don't know in advance which disks will be 4Kn, so the Ignition config will always create a BIOS boot partition on x86_64, but the Dracut glue will automatically skip copying it (and the boot sector) if there's no source partition to copy from.
```
dustymabe: @jlebon and I talked in the past and we seemed to agree that for simplicity sake it would be nice if the partition scheme even for 4k disks matched the non-4k disks. i.e. we were proposing at the time that we have the same number of partitions for both types even if there is a partition that isn't used at all 4k.

WDYT?

bgilbert: If folks are up for it, yeah, I'd prefer that we reduce the number of distinct partition layouts.  :+1:

os/20201103-full-disk-raid.md
Prashanth684: yes zipl would need to be run inrrespective.

os/20201103-full-disk-raid.md
Prashanth684: not sure if the issue is being actively tracked. i will talk to some IBM folks to see if they are aware of anything.

os/20201103-full-disk-raid.md
```diff
@@ -0,0 +1,208 @@
...
+
+Users should only create replicated volumes using the official FCC sugar.  However, each piece of the replication logic (root partition, boot partition, ESP, and BIOS-BOOT + MBR) functions independently, and an advanced user can create nonstandard partition configurations (for example, RAID-1 boot with RAID-5 root) by manually configuring partitions in their FCC.
+
+## FCC sugar
+
+```yaml
+variant: fcos
+version: 1.3.0
+storage:
+  boot_raid:
```
jlebon: Should this be `boot_raid1`?

os/20201103-full-disk-raid.md
```diff
@@ -0,0 +1,208 @@
...
+      - /dev/vda
+      - /dev/vdb
+    # Optional, defaults to x86_64
+    layout: x86_64
+    # Optional, defaults to 0 (entire disk)
+    # Do we want this?
+    root_size_mib: 8192
+    # Optional, defaults to xfs
+    # Do we want this?
+    root_format: xfs
```
jlebon: Is the thinking with these that there's no easy way for users to specify it themselves via `storage.disks.partitions[]`? (Because IIUC, they'd have to know the label names the sugar generates.)

os/20201103-full-disk-raid.md
```diff
@@ -0,0 +1,208 @@
...
+
+## Boot filesystem
+
+Similar to the root filesystem, the Dracut glue will save and restore the contents of `/boot` whenever the config specifies a filesystem labeled `boot` with `wipeFilesystem` set to `true`.
+
+GRUB includes a module for reading MD-RAID volumes.  However, the shipped boot sector is hardcoded to set `prefix` to the first boot disk, and we want to replicate the boot sector via a bit-for-bit copy (see below).  Therefore, we'll create the RAID volume with metadata format 1.0, which puts the RAID superblock at the end of the partition.  This allows GRUB to treat individual `/boot` replicas as independent filesystems, provided that the RAID module is _not_ preloaded into the BIOS boot partition.  (It currently is not.)
+
+`/boot` replication prevents the use of the grubenv mechanism.  If GRUB treats the replicas as independent filesystems, grubenv will cause the RAID to desynchronize.  Alternatively, if we were to use the GRUB RAID module, GRUB would [disable grubenv support](https://www.gnu.org/software/grub/manual/grub/grub.html#Environment-block).
```
jlebon, line 339: It'd be helpful to see what RHEL/Fedora does here. Have you had a chance to look there?

os/20201103-full-disk-raid.md
```diff
@@ -0,0 +1,208 @@
...
+
+There's no BIOS boot partition in 4Kn images.  We don't know in advance which disks will be 4Kn, so the Ignition config will always create a BIOS boot partition on x86_64, but the Dracut glue will automatically skip copying it (and the boot sector) if there's no source partition to copy from.
+
+## PReP partition
+
+For ppc64le, we'll create independent replicas of the PReP partition, similar to the BIOS boot partition.  We'll need to confirm with multi-arch folks that this will work.
+
+## Ignition changes
+
+The Dracut glue can detect the requisite Ignition config stanzas using `jq`, but that's brittle and we're already doing a lot of it in the initrd.  Ideally Ignition will provide command-line arguments for executing relevant (hard-coded) queries against the cached Ignition config.
```
jlebon: I know what you mean, but to clarify maybe let's add a specific example here?

os/20201103-full-disk-raid.md
```diff
@@ -0,0 +1,208 @@
...
+
+## Architecture support
+
+We'll need to get feedback from the multi-arch team for aarch64 and ppc64le.  ppc64le will need an alternate `layout` definition that includes the PReP partition.
+
+For s390x, Ignition doesn't yet support DASD partitioning, and we'd need provisions for rerunning zipl from the initramfs, if RAID support is even possible at all.  Each DASD is connected via a particular bus identifier, which might need to be encoded into each respective copy of the bootloader.  We may need to declare full-disk RAID out of scope for s390x, at least for 4.7.
+
+## Configuring boot disk RAID in OCP
+
+Since FCCs aren't yet used directly in MachineConfigs and FCCT is not shipped with OCP, we'll need a workaround to allow FCC sugar to be used in 4.7.  The MCO could:
```
cgwalters: xref https://github.com/openshift/enhancements/pull/525

I'm uncertain though that the MCO is required here; I think in some cases this may need to be per-machine state and actually be part of the pointer config, which argues for having `oc ignition fcct`/`oc ignition rhcct` or something, see also https://github.com/openshift/oc/pull/628

IOW wherever we say `fcct` we can say `oc ignition fcct` for OKD/FCOS too or so.

Alternatively we could add this functionality directly to MachineConfig, something like an even simpler
```
storage:
  boot_raid:
    - /dev/vda
    - /dev/vdb
```

Basically this path is making MachineConfig a peer of FCC.

(Or we could just punt this section for a second phase)

cgwalters (approved):
Overall seems sane

os/20201103-full-disk-raid.md
bgilbert: It's not completely `raid` anyway.  Updated to `boot_mirror`.

os/20201103-full-disk-raid.md
```diff
@@ -0,0 +1,208 @@
...
+
+`/boot` replication prevents the use of the grubenv mechanism.  If GRUB treats the replicas as independent filesystems, grubenv will cause the RAID to desynchronize.  Alternatively, if we were to use the GRUB RAID module, GRUB would [disable grubenv support](https://www.gnu.org/software/grub/manual/grub/grub.html#Environment-block).
+
+## EFI System Partition
+
+The Dracut glue will save the contents of the ESP filesystem whenever the config specifies a partition with the ESP type GUID, and will restore those contents to _every_ partition with that type GUID.
+
+Thus the ESP won't be RAIDed; we'll create multiple independent filesystems with identical directory trees.  Keeping each replica independent avoids RAID desynchronization if the firmware chooses to modify the ESP, and makes sense because CoreOS does not modify the ESP after installation.  Since there will no longer be a single canonical ESP, we'll no longer automount `/boot/efi` in the running system.
```
travier: This is currently planned as https://github.com/coreos/fedora-coreos-tracker/issues/652

travier (approved):

os/20201103-full-disk-raid.md
bgilbert: :+1: An issue link would be helpful for future reference.

os/20201103-full-disk-raid.md
bgilbert: It doesn't seem great to expect users to know the labels, but we could document them.  And users could still use the "create a subsequent partition with a specified starting offset" trick.

os/20201103-full-disk-raid.md
bgilbert: Neither CentOS 8 nor Fedora 33 provide templates in Anaconda for setting up RAID.  When doing it by hand in the Anaconda partitioner, it's possible to set up RAIDed `/boot` and `/boot/efi`.  In that case, Anaconda automatically uses a RAID 1.0 superblock for the ESP but not for `/boot`.

Also of note, Anaconda doesn't provide a mechanism for creating partitions inside a RAID device.

bgilbert:
Updated to address most of the comments.

os/20201103-full-disk-raid.md
```diff
@@ -0,0 +1,225 @@
...
+
+Thus the ESP won't be RAIDed; we'll create multiple independent filesystems with identical directory trees.  Keeping each replica independent avoids RAID desynchronization if the firmware chooses to modify the ESP, and makes sense because CoreOS does not modify the ESP after installation.  Since there will no longer be a single canonical ESP, we'll no longer automount `/boot/efi` in the running system (see [coreos/fedora-coreos-tracker#652](https://github.com/coreos/fedora-coreos-tracker/issues/652)).
+
+When updating the bootloader, bootupd will need to find each partition with an ESP type GUID and update it independently.  This assumes that all ESPs on the system are controlled by CoreOS.
+
+We could conditionalize the copy logic to allow a missing source ESP.  Per the usual Ignition philosophy, it'd be better to fail outright if there is no source ESP, but that'd conflict with the [proposal to drop the ESP in AWS images](https://github.com/coreos/fedora-coreos-config/pull/407).
```
arithx: Because we're specifying layout inside of the FCC we could add an AWS specific layout that doesn't have the ESP so I think we're covered no?

os/20201103-full-disk-raid.md
bgilbert: Since s390x is being declared out of scope for now, we're going to stick with partition labels.  s390x, if added, will use a separate config template which can use partition UUIDs if still needed.

bgilbert:
Updated based on refinement session.  PTAL.

Based on a discussion with @arithx, the FCC sugar now provides a unified mechanism for configuring LUKS root, mirrored boot disks, or both.  In an effort to keep the sugar as simple as possible, I've dropped the sugared properties for setting the root filesystem type and partition size for now.  We can always re-add them later, and/or advanced users can use overrides or avoid the sugar entirely.

bgilbert:
Moved `boot_device` sugar to the top level per an OOB discussion with @arithx.

bgilbert:
Moved `layout` directly under `boot_device` to handle the case that some platform/architecture requires different handling for LUKS.

bgilbert:
Okay, made a final set of updates to reflect the current plan.

PR #6, opened 2021-10-18
s390x: SecureExecution enablement

nikita-dubrovskii:
Signed-off-by: Nikita Dubrovskii <nikita@linux.ibm.com>


os/20211014-s390x-secure-execution.md
```diff
@@ -0,0 +1,191 @@
...
+<memballoon model='none'/>
+```
+
+Once this done, `kernel`, `cmdline` and `initramdisk` could be encrypted:
+```
+# genprotimg -i /path/to/vmlinuz-xyz.s390x -r /path/to/initramfs-xyz.s390x.img -p /path/to/cmdline -k hostkey -o secure --no-verify
+```
+
+The fact that `secure-vmlinuz` boots means that SE is enabled and Linux is secured.
+Right now we cannot check it via `/proc/smth`. To indicate that SE is on, we add dummy karg.
```
julianwiedmann: > Right now we cannot check it via `/proc/smth`.

What about `/sys/firmware/uv/prot_virt_guest` ?

os/20211014-s390x-secure-execution.md
```diff
@@ -0,0 +1,191 @@
...
+the host can access the memory and data of guest applications freely, leading to the
+potential for malicious software to be proliferated throughout the entire system.
+Isolation between host and guest environments is necessary to help prevent system compromise.
+Users can enable Secure Execution feature on IBM zKVM Host and Guest VMs to isolate them from each other.
+
+That means:
+1. An encrypted Linux image (kernel + ramdisk + cmdline) is created using the host public key and a customer-specific key. Since the encryption keys are stored on the IBM Z hardware and firmware, the encrypted image can only be executed in a virtual machine on the host(s) it has been prepared for, and the image can’t be decrypted or tampered with outside of the designated host(s).
+2. zKVM host cannot acces memory of secured guest VM
+3. Guest VMs cannot access memory of other secured guest VMs
+4. If anyone tries to access memeory of secured guest VM, than this gets blocked on HW level
```
JAORMX: s/memeory/memory/

os/20211014-s390x-secure-execution.md
```diff
@@ -0,0 +1,191 @@
...
+> **_NOTE:_**
+`Hostkey` is generated by IBM and is stored somewhere in HW, the key is unique for each IBM Z mainframe. Image signed with a key could be used only on machine with the same key.
+
+> **_NOTE:_**
+SE doesn't require guest's disk(s) to be encrypted, however it's recomended to enable LUKS and save the key to `initramdisk`. After generation of `secure-vmlinuz` old `vmlinuz` and `initramdisk` could be erased from filesystem.
+
+
+# High-level overview
+
+SE is supported only startuing with z15 and only for zKVM guests, so SE shouldn't be used with zVM.
```
JAORMX: s/startuing/starting/

os/20211014-s390x-secure-execution.md
```diff
@@ -0,0 +1,191 @@
...
+built into the IBM z15™ and LinuxONE III generation systems. It is designed to
+provide scalable isolation for individual workloads to help protect them from not
+only external attacks, but also insider threats. In traditional x86 ring architectures,
+the host can access the memory and data of guest applications freely, leading to the
+potential for malicious software to be proliferated throughout the entire system.
+Isolation between host and guest environments is necessary to help prevent system compromise.
+Users can enable Secure Execution feature on IBM zKVM Host and Guest VMs to isolate them from each other.
+
+That means:
+1. An encrypted Linux image (kernel + ramdisk + cmdline) is created using the host public key and a customer-specific key. Since the encryption keys are stored on the IBM Z hardware and firmware, the encrypted image can only be executed in a virtual machine on the host(s) it has been prepared for, and the image can’t be decrypted or tampered with outside of the designated host(s).
```
JAORMX: Currently, the Machine Config Operator allows for configuring settings in the linux cmdline via MachineConfig. e.g.

```
apiVersion: machineconfiguration.openshift.io/v1
kind: MachineConfig
metadata:
  ...
spec:
  config:
    ignition:
      version: 3.1.0
  kernelArguments:
    - audit=1
```

Does this make it so that the kernel arguments are no longer configurable? or are there plans to make it so that changes to the cmdline through the MCO are re-encrypted with this mechanism?

JAORMX (changes requested):

os/20211014-s390x-secure-execution.md
nikita-dubrovskii: Any change of kargs on s390x reuqires `zipl`, so we'd have to re-encrypt everything

os/20211014-s390x-secure-execution.md
nikita-dubrovskii: Thx!

os/20211014-s390x-secure-execution.md
JAORMX: @nikita-dubrovskii is re-encrypting in scope for the current implementation? or will that come in a further iteration?

nikita-dubrovskii: i don't know how MCO modifies kernel's cmdline, but if it's done via `rpm-ostree kargs`, than re-encryption would be done by `ostree`

os/20211014-s390x-secure-execution.md
```diff
@@ -0,0 +1,190 @@
...
```
cgwalters: This concept of "single signed blob that is kernel+initramfs+cmdline" is used in other places too.  `sd-boot` semi-emphasizes it.

I think we may end up wanting to at least make ostree more aware of this.  In particular, there are parallels to the EFI case around teaching ostree that it needs to do something extra here - in the EFI case, writing binaries to the ESP and not `/boot`.

os/20211014-s390x-secure-execution.md
```diff
@@ -0,0 +1,190 @@
...
+2. zKVM host cannot acces memory of secured guest VM
```
cgwalters:
```suggestion
2. zKVM host cannot access memory of secured guest VM
3. ```

File: os/20211014-s390x-secure-execution.md [hunk 7]
```diff
@@ -0,0 +1,190 @@
...
+3. Guest VMs cannot access memory of other secured guest VMs
+4. If anyone tries to access memory of secured guest VM, than this gets blocked on HW level
+
+## Specific attacks being addressed
+- protect guests against the system
+- protect guests against a Linux Admin
+- protect services in KVM guests against each other
+
+## Specific attacks not being addressed
+- build/enablement must happen in a trusted environment, so we don't consider `hacking` during Ignition
```
cgwalters:
```suggestion
- This does not attempt to secure initial system provisioning (Ignition)
```

os/20211014-s390x-secure-execution.md
```diff
@@ -0,0 +1,190 @@
...
+zipl
+reboot
+```
+
+# Design
+
+## Ignition
+
+We don't need to have any special section for SE. User just defines a `file` with pre-defined name
+`/etc/ibm-z-hostkey`:
```
cgwalters: Is this file read by `genprotimg` by default?

os/20211014-s390x-secure-execution.md
```diff
@@ -0,0 +1,190 @@
...
+
+## Dracut
+We add a new dracut module to CoreOS that looks for the presence of `/run/s390-secure-encryption-luks.key`. If found, it mounts the `/boot` partition and runs `ostree s390-genprotimage -k /sysroot/etc/genprotimage.key`.
+
+This would take the currently configured kernel/initramfs and run `genprotimage` as well as `zipl`.
+
+## ostree
+Additionally, the OSTree bootloader backend is configured by this dracut module:
+```
+sed -i -e s,bootloader=zipl,bootloader=s390-secureexec /sysroot/ostree/repo/config
```
cgwalters: We can probably auto-detect this case too.

os/20211014-s390x-secure-execution.md
```diff
@@ -0,0 +1,190 @@
...
+
+# High-level overview
+
+SE is supported only starting with z15 and only for zKVM guests, so SE shouldn't be used with zVM.
+zKVM host should be configured before turning SE on:
+- kernel cmdline should have `prot_virt=1 swiotlb=262144`
+- `hostkey` (i.e. HKD-ABCD-XXYYZZ.crt) must be downloaded for each particular machine
+- only production keys could be verified
+
+After guest VM is installed, it's config file should be modified:
```
cgwalters: Perhaps something like:

> The following are example [libvirt](https://libvirt.org/) XML stanzas:

?

cgwalters (commented):
I think it might be fair to summarize the basics of this as:

"Enable a rough equivalent of the effect of TPM-bound LUKS on z15, by storing the LUKS key in the initramfs, which is encrypted by the host" ?

os/20211014-s390x-secure-execution.md
```diff
@@ -0,0 +1,190 @@
+# IBM Z Secure Execution
+---
+
+# Theory of operation
+
+IBM Secure Execution for Linux® is a hardware-based security technology that is
+built into the IBM z15™ and LinuxONE III generation systems. It is designed to
+provide scalable isolation for individual workloads to help protect them from not
```
holgwolf: it is isolation among Guest so that nobody can break memory boundaries among two qemu processes. it is workload somewhat but specifically packed in KVM guests.

os/20211014-s390x-secure-execution.md
```diff
@@ -0,0 +1,190 @@
+# IBM Z Secure Execution
+---
```
julianwiedmann, line 2: Some high-level thoughts, will need to dig into the details later.

1. I think it's worth pointing at some high-level docs to provide more context. For instance https://www.ibm.com/docs/en/linux-on-systems?topic=virtualization-introducing-secure-execution-linux
2. SE prevents the use of some features (most notably passthrough of host devices). Do we have a dependency on any of those features - CI environments that use vfio-ccw or such?
3. We'll need to consider the additional memory requirements for SE in our sizing recommendations.
4. The docs suggest to blacklist the `virtio-rng` module in the SE guest.
5. Do we need to collect/wire up any additional debug output for SE issues?

os/20211014-s390x-secure-execution.md
```diff
@@ -0,0 +1,190 @@
...
+# Alternative design
+
+## Ignition
+
+We add to new `3.x.y-experimental` special option and step for SE:
+
+```yaml
+storage:
+  s390SecureEncryption:
+    device: /dev/disk/by-partlabel/root
```
julianwiedmann: Are we at risk of working with a non-GPT partition, eg. virt-blk backed by a DASD? Then I wouldn't expect `by-partlabel` to work.

os/20211014-s390x-secure-execution.md
holgwolf: Comments on 2 points
no 2 is a documented restriction in point in time - I do not see any dependencies. This is for me part of the docs. 
no 3 will need to be considered in the docs but may not depend here on the design.

os/20211014-s390x-secure-execution.md
nikita-dubrovskii: no, but i can ask` s390-tools` team about having some `defaults`

os/20211014-s390x-secure-execution.md
nikita-dubrovskii: > The docs suggest to blacklist the virtio-rng module in the SE guest.

haven't seen that module on running FCOS VM, so skipped it here

os/20211014-s390x-secure-execution.md
```diff
@@ -0,0 +1,190 @@
...
+
+mount -o remount,rw /boot
+for config in /boot/loader/entries/*.conf; do
+    opts=$(grep options $config | cut -d' ' -f2-)
+    commit=$(grep linux $config | cut -d' ' -f2- | cut -d'/' -f3)
+    kernel=$(find /boot/ostree/$commit/ -name "vmlinuz-*")
+    initrd=$(find /boot/ostree/$commit/ -name "initramfs-*")
+    secure=/ostree/$commit/secure
+    echo "$opts swiotlb=262144 secured" > /tmp/cmdline
+    genprotimg -i $kernel -r $initrd -p /tmp/cmdline -k $key --no-verify -o /boot/$secure
```
julianwiedmann: Is `--no-verify` ok for production usage? Or do we also need to consider where the root CA certificate comes from?

os/20211014-s390x-secure-execution.md
nikita-dubrovskii: from Slack chat with @julianwiedmann:
 
>https://bugzilla.linux.ibm.com/show_bug.cgi?id=194983
>Sounds like works-as-designed, modern guests are meant to prefer s390-trng over virtio-rng.

os/20211014-s390x-secure-execution.md
```diff
@@ -0,0 +1,201 @@
...
+built into the IBM z15™ and LinuxONE III generation systems. It is designed to
+provide scalable isolation for individual workloads to help protect them from not
+only external attacks, but also insider threats. In traditional x86 ring architectures,
+the host can access the memory and data of guest applications freely, leading to the
+potential for malicious software to be proliferated throughout the entire system.
+Isolation between host and guest environments is necessary to help prevent system compromise.
+Users can enable Secure Execution feature on IBM zKVM Host and Guest VMs to isolate them from each other.
+
+That means:
+1. An encrypted Linux image, called `sd-boot`, (kernel + ramdisk + cmdline) is created using the host public key and a customer-specific key. Since the encryption keys are stored on the IBM Z hardware and firmware, the encrypted image can only be executed in a virtual machine on the host(s) it has been prepared for, and the image can’t be decrypted or tampered with outside of the designated host(s).
```
bgilbert: What does the customer-specific key do?  Does the Z firmware manage it also?

os/20211014-s390x-secure-execution.md
```diff
@@ -0,0 +1,201 @@
...
+2. zKVM host cannot access memory of secured guest VM
+3. Guest VMs cannot access memory of other secured guest VMs
+4. If anyone tries to access memory of secured guest VM, than this gets blocked on HW level
+
+## Specific attacks being addressed
+- protect guests against the system
+- protect guests against a Linux Admin
```
bgilbert: To be clear, I assume this is a Linux admin on the host, not the guest.

os/20211014-s390x-secure-execution.md
```diff
@@ -0,0 +1,201 @@
...
+- access to contents of guest's hard-disk image (qcow2, raw, ...)
+
+> **_NOTE:_**
+`Hostkey` is generated by IBM and is stored somewhere in HW, the key is unique for each IBM Z mainframe. Image signed with a key could be used only on machine with the same key.
+
+> **_NOTE:_**
+SE doesn't require guest's disk(s) to be encrypted, however it's recomended to enable LUKS and save the key to `initramdisk`. After generation of `sd-boot` old `vmlinuz` and `initramdisk` could be erased from filesystem.
+
+
+# High-level overview
```
bgilbert, line 56: This section is useful, but it seems like more of a hacking guide than a high-level overview.  Could we also have a section that states what elements of the system are initially trusted, and explains the process of extending and maintaining trust outward from that initial component?

For example: the customer owns a machine which is fully trusted, and is used to generate the original sd-boot image.  Then that image is sent to the cloud, which verifies that the kernel and initramfs are unmodified.  The Ignition config is then fetched from [somewhere] and we verify that it hasn't been modified by doing [X].  We verify that the boot disk is unmodified by doing [Y], and boot into the system.  Then, when installing updates, we do [Z] which ensures that integrity is maintained when booting into the new system.

We'll also eventually have to consider some other factors, such as:
- How can the cluster know that the host didn't simply boot an RHCOS image with SE disabled?
- Are we concerned about the host forcing downgrades to previous OS releases that might contain security holes?  If so, how do we mitigate that?

os/20211014-s390x-secure-execution.md
```diff
@@ -0,0 +1,201 @@
...
+
+We don't need to have any special section for SE. User just defines a `file` with pre-defined name `/etc/ibm-z-hostkey-N` where `N` is a key number, multiple keys could be used:
+```yaml
+storage:
+  files:
+    -
+      path: /etc/ibm-z-hostkey-1
+      overwrite: true
+      contents:
+        source: http://url.com/key1
```
bgilbert: This might have been meant as a placeholder, but the URL implies that we're fetching the hostkey without any TLS protection or hash checks.  Is that safe and intended?

os/20211014-s390x-secure-execution.md
```diff
@@ -0,0 +1,201 @@
...
+    rm /tmp/cmdline
+    sed -i '/linux\|options\|initrd */d' $config
+    echo "linux $secure" >> $config
+    echo "options $opts" >> $config
+done
+zipl
+reboot
+```
+
+# Design
```
bgilbert: The rest of this document gives a lot of specific implementation details, but doesn't explain why they need to exist.  The details are useful, but please also explain what each part of the design is trying to accomplish.

os/20211014-s390x-secure-execution.md
```diff
@@ -0,0 +1,201 @@
...
+2. zKVM host cannot access memory of secured guest VM
+3. Guest VMs cannot access memory of other secured guest VMs
+4. If anyone tries to access memory of secured guest VM, than this gets blocked on HW level
+
+## Specific attacks being addressed
+- protect guests against the system
+- protect guests against a Linux Admin
+- protect services in KVM guests against each other
+
+## Specific attacks not being addressed
```
bgilbert, line 44: Okay, so as I understand this:
- We're not securing provisioning, so we're okay with the host accessing/modifying guest memory on first boot and potentially installing a rootkit
- We're not requiring disk encryption, so we're okay with the host reading guest data
- We're not requiring authenticated disk encryption, so we're okay with the host compromising software on the guest's disk

In what way are we protecting guests against the system, a Linux admin, or each other?

os/20211014-s390x-secure-execution.md
hbrueckner: @bgilbert This customer-specific key might be used to encrypt  data (by UV) to be sent back to the customer. The customer-specific key is part of the SE header being encrypted with the one or more host public keys.

os/20211014-s390x-secure-execution.md
```diff
@@ -0,0 +1,201 @@
+# IBM Z Secure Execution
```
hbrueckner: Please use correct wording: "IBM Secure Execution for Linux"

os/20211014-s390x-secure-execution.md
```diff
@@ -0,0 +1,201 @@
...
+That means:
+1. An encrypted Linux image, called `sd-boot`, (kernel + ramdisk + cmdline) is created using the host public key and a customer-specific key. Since the encryption keys are stored on the IBM Z hardware and firmware, the encrypted image can only be executed in a virtual machine on the host(s) it has been prepared for, and the image can’t be decrypted or tampered with outside of the designated host(s).
+2. zKVM host cannot access memory of secured guest VM
+3. Guest VMs cannot access memory of other secured guest VMs
+4. If anyone tries to access memory of secured guest VM, than this gets blocked on HW level
+
+## Specific attacks being addressed
+- protect guests against the system
+- protect guests against a Linux Admin
+- protect services in KVM guests against each other
```
hbrueckner: You need to be clear what protecting services mean here. You can protect the data the services use (within a secured guest). You cannot protect against DoS attacks or vulnerabilities in the service itself.

ashcrow:
I see some good discussion here with some clarifications and some open questions. Could we get some of those items refined back into the enhancement for further review and clarification or is that happening in another location?

os/20211014-s390x-secure-execution.md
bgilbert: Oh, interesting.  Does that mean the VM could encrypt a nonce using the customer-specific key to prove that it was booted via SE?  Can the customer-specific key be chosen arbitrarily when the sd-boot image is generated?

hbrueckner: The customer-specific key is actually the key the customer owns and the customer generated the sd-boot image (on trusted hardware).

os/20211014-s390x-secure-execution.md
nikita-dubrovskii: that's a placeholder. I expect that key comes as base64 or via https

os/20211014-s390x-secure-execution.md
nikita-dubrovskii: as i understand it - main goal is to protect guest's RAM, disk encryption is a bit aside (but recommended)
@hbrueckner , @holgwolf wdyt?

os/20211014-s390x-secure-execution.md
nikita-dubrovskii: >How can the cluster know that the host didn't simply boot an RHCOS image with SE disabled?

right now only via ssh-ing to the guest

>Are we concerned about the host forcing downgrades to previous OS releases that might contain security holes?

Downgrading with disabling SE shouldn't be supported. but downgrading itself i guess should be OK.
@hbrueckner ?

hbrueckner: > > How can the cluster know that the host didn't simply boot an RHCOS image with SE disabled?
> 
> right now only via ssh-ing to the guest

Well, my understanding is that the entire RHOCP cluster entirely within the SE boundary (multiple SE guests). That would mean, a "regular" RHCOS image will never be part of the cluster. Secondly, no SE, no SE-secured guest, no RHOCP node.
 
> > Are we concerned about the host forcing downgrades to previous OS releases that might contain security holes?
> 
> Downgrading with disabling SE shouldn't be supported. but downgrading itself i guess should be OK. @hbrueckner ?

Some parts might be covered with attestation.. as the goal of SE is to protect against data access from the hypervisor, this should then covered for OS downgrades as well.

cgwalters:
I just wanted to add an overall comment that I see the value here.  Protecting against an attacker with full root on the hypervisor is hard, but I think there are a number of historical vulnerabilities that would be mitigated by this.  Things like [this Xen CVE](https://www.cvedetails.com/cve/CVE-2021-28701/) and I think there have been similar for KVM.

os/20211014-s390x-secure-execution.md
```diff
@@ -0,0 +1,215 @@
...
+3. Guest VMs cannot access memory of other secured guest VMs
+4. If anyone tries to access memory of secured guest VM, than this gets blocked on HW level
+
+## Specific attacks being addressed
+- protect guests against the system
+- protect guests against a Linux Admin (on the host)
+- protect the data services use within a secured guests from host admin and/or other guests
+
+## Specific attacks not being addressed
+- this does not attempt to secure initial system provisioning (Ignition)
```
julianwiedmann: Just to be clear - you say that firstboot will be done without SE? It's not obvious to me why that's ok / desired.

My expectation would have been that the user either manually turns the fcos install artifacts into an SE bootimage, or coreos-installer (?) learns how to build the SE bootimage for installation.

os/20211014-s390x-secure-execution.md
bgilbert: Okay, let's adjust to `https` then.

bgilbert: (whoops, sorry, missed the update)

os/20211014-s390x-secure-execution.md
```diff
@@ -0,0 +1,157 @@
...
+That means, that user won't notice any difference, both mentioned components would automatically regenerate `sd-boot` image and
+install it with `zipl` whenever required.
+
+## Ignition Protection
+
+Ignition config will be built and when saved encrypted with a customer's public key.
+The private key is part of the sd-boot image and is generated by RH/IBM during qcow2 image build. The public key is part of the build metadata.
+During firstboot the Ignition config gets decrypted and the key gets removed immediately after it. This could be done with a
+
+## Emergecny Shell
```
bgilbert: `Emergency`

os/20211014-s390x-secure-execution.md
```diff
@@ -0,0 +1,157 @@
...
+built into the IBM z15™ and LinuxONE III generation systems. It is designed to
+provide scalable isolation for individual workloads to help protect them from not
+only external attacks, but also insider threats. In traditional x86 ring architectures,
+the host can access the memory and data of guest applications freely, leading to the
+potential for malicious software to be proliferated throughout the entire system.
+Isolation between host and guest environments is necessary to help prevent system compromise.
+Users can enable Secure Execution feature on IBM zKVM Host and Guest VMs to isolate them from each other.
+
+That means:
+1. An encrypted Linux image, called `sd-boot`, (kernel + initramfs + cmdline) is created using the host public key and a customer-specific key. Since the encryption keys are stored on the IBM Z hardware and firmware, the encrypted image can only be executed in a virtual machine on the host(s) it has been prepared for, and the image can’t be decrypted or tampered with outside of the designated host(s).
```
bgilbert: Are we still using a customer-specific key?

os/20211014-s390x-secure-execution.md
```diff
@@ -0,0 +1,157 @@
...
+4. Guest VMs cannot access memory of other secured guest VMs
+5. If anyone tries to access memory of secured guest VM, than this gets blocked on HW level
+
+## Specific attacks being addressed
+- protect guests against the system
+- protect guests against a Linux Admin (on the host)
+- protect the data services use within a secured guests from host admin and/or other guests
+
+## Specific attacks not being addressed
+- any attack over Internet on some serivce(s) running within Guest
```
bgilbert: `service(s)`

os/20211014-s390x-secure-execution.md
```diff
@@ -0,0 +1,157 @@
...
+- zKVM's admin, who tries to hack OCP with it's own CoreOS guest and/or ignition config
+- usage of the fake ignition config for guest provisioning
+
+> **_NOTE:_**
+`Hostkey` is generated by IBM and is stored somewhere in HW, the key is unique for each IBM Z mainframe.
+Image signed with a key (or set of keys) could be used only on machine with the identical key.
+
+> **_NOTE:_**
+Only production hostkeys could be verified
```
bgilbert: I'm not entirely sure what this means.  Another sentence or two would be useful.

os/20211014-s390x-secure-execution.md
```diff
@@ -0,0 +1,157 @@
...
+
+
+# High-level overview
+
+The goal of SecureExecution is to provide customer the highest level of isolation and security for each OCP node.
+To achive that new type of artifact for s390x would be available to download and use - SecureExecution Ready Qcow2 Image:
+ - Image comes with `sd-boot` generated using universal key from IBM. Key itself is not known to public and used by RH/IBM during build of RHCOS
+ - Image couldn't be modified, by anyone except `ignition` during lifetime
+ - It's won't possible to inspect and modify contents of image's `/` and `/boot` filesystems
```
bgilbert: It'll still be possible to `oc debug` or SSH to the node and modify the filesystem, right?

os/20211014-s390x-secure-execution.md
```diff
@@ -0,0 +1,157 @@
...
+ - It's only possible to run this image on zKVM host with enabled SecureExecution support
+ - It's not possible to disable SecureExecution once it was enabled
+
+
+ Also it would be possible to enable SecureExecution during guest provisioning on regular qcow2 image.
```
bgilbert: Do we plan to support this?

os/20211014-s390x-secure-execution.md
```diff
@@ -0,0 +1,157 @@
...
+
+
+##  Build
+ To build SecureExecution Ready image new command is added to `coreos-assembler`:
+ ```
+ cosa buildextend-se --hostkey UNIVERSAL_KEY.crt
+ ```
+ Which is similar to `cosa buildextend-qemu` and also generates qcow2 image.
+
+The qcow2 image comes with `sd-boot` isntalled into its own partition, and LUKS2 pre-encrypted boot and rootfs filesystems.
```
bgilbert: `installed`

os/20211014-s390x-secure-execution.md
```diff
@@ -0,0 +1,157 @@
...
+  `-crypt_rootfs root         crypt /sysroot
+
+```
+
+- `se`   - This is a new partition `ext4`, where we store a `sd-boot` encrypted image
+- `boot` - boot partition is the same as one on reqular qcow2 image, additionally it's encrypted with random LUKS key during CoreOS build process
+- `root` - rootfs partition is almost the same as one on reqular qcow2 image, additionally it's encrypted with random LUKS key during CoreOS build process
+and contains `/etc/crypttab`, `/etc/luks/root` and `/etc/luks/boot` files.
+
+The reason why `boot` has to be encrypted is that without encryption hacker is able to mount it and modify its contents, for example add new kargs to BLS config file(s), so during system's lifetime that modification could affect `sd-boot` image.
```
bgilbert: To be clear, encryption alone isn't sufficient for that purpose; we also need to enable LUKS integrity protection.

os/20211014-s390x-secure-execution.md
```diff
@@ -0,0 +1,157 @@
...
+
+Despite `zipl` is well known and used by CoreOS, it couldn't be executed same way as before. When system runs in SecureExecution, `ostree` and `rdcore` automatically detect that and perform all required steps.
+That means, that user won't notice any difference, both mentioned components would automatically regenerate `sd-boot` image and
+install it with `zipl` whenever required.
+
+## Ignition Protection
+
+Ignition config will be built and when saved encrypted with a customer's public key.
+The private key is part of the sd-boot image and is generated by RH/IBM during qcow2 image build. The public key is part of the build metadata.
+During firstboot the Ignition config gets decrypted and the key gets removed immediately after it. This could be done with a
```
bgilbert: This section is incomplete.  I no longer recall the details; is this still included in the plan?

os/20211014-s390x-secure-execution.md
```diff
@@ -0,0 +1,157 @@
...
+
+## Emergecny Shell
+
+Before entering the initramfs emergency shell, the private key is removed to avoid leaking it.
```
bgilbert: Does this refer to the private key mentioned above for Ignition config encryption, or to the disk encryption key, or both?

os/20211014-s390x-secure-execution.md
nikita-dubrovskii: `host public key` (i.e. universal key) is used during `cosa build`, and indeed it's not public at all . than during `firstboot` we regenerate `sd-boot` with customer's key(s)

os/20211014-s390x-secure-execution.md
nikita-dubrovskii: yes

os/20211014-s390x-secure-execution.md
nikita-dubrovskii: not sure, will remove it

os/20211014-s390x-secure-execution.md
nikita-dubrovskii: as step4. But this is also to be discussed later.

os/20211014-s390x-secure-execution.md
nikita-dubrovskii: here it's about ignition key

os/20211014-s390x-secure-execution.md
```diff
@@ -0,0 +1,157 @@
...
+potential for malicious software to be proliferated throughout the entire system.
+Isolation between host and guest environments is necessary to help prevent system compromise.
+Users can enable Secure Execution feature on IBM zKVM Host and Guest VMs to isolate them from each other.
+
+That means:
+1. An encrypted Linux image, called `sd-boot`, (kernel + initramfs + cmdline) is created using the host public key and a customer-specific key. Since the encryption keys are stored on the IBM Z hardware and firmware, the encrypted image can only be executed in a virtual machine on the host(s) it has been prepared for, and the image can’t be decrypted or tampered with outside of the designated host(s).
+2. kernel, initramfs and cmdline couldn't be modified without regeneration of `sd-boot` image
+3. zKVM host cannot access memory of secured guest VM
+4. Guest VMs cannot access memory of other secured guest VMs
+5. If anyone tries to access memory of secured guest VM, than this gets blocked on HW level
```
dustymabe:
```suggestion
5. If anyone tries to access memory of secured guest VM, the attempt will get blocked at the HW level
```

os/20211014-s390x-secure-execution.md
```diff
@@ -0,0 +1,157 @@
...
+- protect the data services use within a secured guests from host admin and/or other guests
+
+## Specific attacks not being addressed
+- any attack over Internet on some service(s) running within Guest
+- zKVM's admin, who tries to hack OCP with it's own CoreOS guest and/or ignition config
+- usage of the fake ignition config for guest provisioning
+
+> **_NOTE:_**
+`Hostkey` is generated by IBM and is stored somewhere in HW, the key is unique for each IBM Z mainframe.
+Image signed with a key (or set of keys) could be used only on machine with the identical key.
```
dustymabe:
```suggestion
> **Note**
> `Hostkey` is generated by IBM and is stored somewhere in HW, the key is unique for each IBM Z mainframe.
> Image signed with a key (or set of keys) could be used only on machine with the identical key.
```

and the same for the next line.. See https://github.com/github/feedback/discussions/16925

os/20211014-s390x-secure-execution.md
```diff
@@ -0,0 +1,157 @@
...
+5. If anyone tries to access memory of secured guest VM, than this gets blocked on HW level
+
+## Specific attacks being addressed
+- protect guests against the system
+- protect guests against a Linux Admin (on the host)
+- protect the data services use within a secured guests from host admin and/or other guests
+
+## Specific attacks not being addressed
+- any attack over Internet on some service(s) running within Guest
+- zKVM's admin, who tries to hack OCP with it's own CoreOS guest and/or ignition config
```
dustymabe: I'd prefer to keep this generic and not mention OCP here (since this is coreos/enhancements and not openshift/enhancements). Can we make the example more generic?

os/20211014-s390x-secure-execution.md
```diff
@@ -0,0 +1,157 @@
...
+
+The goal of SecureExecution is to provide customer the highest level of isolation and security for each OCP node.
+To achive that new type of artifact for s390x would be available to download and use - SecureExecution Ready Qcow2 Image:
+ - Image comes with `sd-boot` generated using universal key from IBM. Key itself is not known to public and used by RH/IBM during build of RHCOS
+ - Image couldn't be modified, by anyone except `ignition` during lifetime
+ - It won't possible to inspect and modify (using `guestfish` or other tools) contents of image's `/` and `/boot` filesystems
+ - It's only possible to run this image on zKVM host with enabled SecureExecution support
+ - It's not possible to disable SecureExecution once it was enabled
+
+To use SecureExecution customer must have hostkeys for hos machine(s) and specify them in `ignition` config.
```
dustymabe:
```suggestion
To use SecureExecution customer must have hostkeys for host machine(s) and specify them in `ignition` config.
```

os/20211014-s390x-secure-execution.md
bgilbert: I'm confused about the "customer keys" part.  My understanding is that we generate a random key for disk encryption, and we use the host keys of the target machines for sd-boot encryption, but that the customer does not supply their own key for the sd-boot image.

os/20211014-s390x-secure-execution.md
bgilbert: Let's clarify that it won't be possible to do these things from outside the VM, then.

os/20211014-s390x-secure-execution.md
nikita-dubrovskii: sorry, i wrote not  clear. so, we have 2 sets of keys - 1 for SE, and 1 for LUKS2. As you said, LUKS keys are random, and customer/user doesn't touch them. But for SE customer has its own key(s), which are listed in `ignition_config.ign` , and these keys are replacing universal hostkey from IBM

os/20211014-s390x-secure-execution.md
nikita-dubrovskii: i've removed this line.

os/20211014-s390x-secure-execution.md
bgilbert: To be clear, those keys are the public keys for the hosts the customer intends to run on, correct?  "Customer-specific key" feels confusing in that context, since the customer doesn't control the corresponding private keys.

os/20211014-s390x-secure-execution.md
```diff
@@ -0,0 +1,156 @@
...
+## Specific attacks not being addressed
+- any attack over Internet on some service(s) running within Guest
+- usage of the fake ignition config for guest provisioning
+
+> **Note:**
+> `Hostkey` is generated by IBM and is stored somewhere in HW, the key is unique for each IBM Z mainframe.
+> Image signed with a key (or set of keys) could be used only on machine with the identical key.
+
+> **Note:**
+> Only production hostkey(s) could be verified (that a host key document is genuine) using [check_hostkeydoc](https://raw.githubusercontent.?com/ibm-s390-tools/s390-tools/master/genprotimg/samples/check_hostkeydoc) tool. In a later version of `genprotimg` this function will
```
bgilbert: Stray `?` in the domain name.

os/20211014-s390x-secure-execution.md
nikita-dubrovskii: yep, fixed

os/20211014-s390x-secure-execution.md
```diff
@@ -0,0 +1,171 @@
...
+That means, that user won't notice any difference, both mentioned components would automatically regenerate `sd-boot` image and
+install it with `zipl` whenever required.
+
+# Implementation steps
+
+## Step 1: CoreOS with existing patches to `libostree` could be configured to perform in SecureExecution mode.
+
+Customer public hostkeys should be placed into `/etc/se-hostkeys/` using name templates `ibm-z-hostkey-N`. SecureExecution is enabled by hands following https://www.ibm.com/docs/en/linux-on-systems?topic=linux-introduction.
+
+## Step 2: CoreOS with is always performing in SecureExecution mode.
```
bgilbert: Would be good to reword the title to clarify that this is only true of the SE image and not the regular one.

os/20211014-s390x-secure-execution.md
```diff
@@ -0,0 +1,171 @@
...
+
+Pre-build SecureExecution-ready qcow2 with LUKS encrypted `boot-` and `rootfs` is available for download. The image contains the `sd-boot` image signed with universal public hostkey, which is valid for all Z15 and newer machines. Customer should add his public hostkeys(s) into `ignition` config.
+
+## Step 3: Reencryption of `boot-` and `rootfs` with random LUKS keys and `integrity` option.
+
+During firstboot, the LUKS keys will be changed to a randomly generated ones. The idea behind this is to set on each host the LUKS keys to a unique keys during firstboot. This should prevent reading the key from the provided pre-build image.
```
bgilbert: If we're deferring integrity protection to step 3, it'd be good to mention it explicitly in the description, since it doesn't just affect re-encryption.  The pristine image will also need to ship with integrity protection enabled.

os/20211014-s390x-secure-execution.md
```diff
@@ -0,0 +1,171 @@
...
+</disk>
+```
+- network interfaces should have `iommu=on` option:
+```xml
+ <interface type='network'>
+      <model type='virtio'/>
+      <driver name='qemu' iommu='on'/>
+</interface>
+```
+- `membaloon` should be disabled:
```
bgilbert: `memballoon`

os/20211014-s390x-secure-execution.md
```diff
@@ -0,0 +1,171 @@
...
+
+# Detailed overview
+
+To achive goals listed above not only a special qcow2 image is generated, but also some additional steps during system's lifetime are required.
+
+
+##  Build
+ To build SecureExecution Ready image new command is added to `coreos-assembler`:
+ ```
+ cosa buildextend-se --hostkey UNIVERSAL_KEY.crt
```
bgilbert: `-secex` now.

os/20211014-s390x-secure-execution.md
nikita-dubrovskii: > The pristine image will also need to ship with integrity protection enabled.

This doesn't support `growfs`

bgilbert: Then we'll need to reprovision the root filesystem rather than using `growfs`.  If we don't enable integrity protection in the pristine image, a malicious host can modify the disk image before it's booted.

os/20211014-s390x-secure-execution.md
```diff
@@ -0,0 +1,190 @@
...
+
+Customer's public hostkey(s) should be placed into `/etc/se-hostkeys/` using name templates `ibm-z-hostkey-N`. Secure Execution is enabled by hands following https://www.ibm.com/docs/en/linux-on-systems?topic=linux-introduction.
+
+## Step 2: New `qemu-secex.qcow2` build artifact for CoreOS with enabled Secure Execution.
+
+`coreos-xyz-qemu-secex.qcow2` image with LUKS encrypted `/boot` and `rootfs` is available for download. This is done by using new `buildextend-secex` `coreos-assembler` command, which:
+ - adds new `/se` partition for `sd-boot`
+ - encrypts `/boot` and `/` with random LUKS keys
+ - stores LUKS keys and `crypttab` config to `/etc`
+ - enables a rough equivalent of the effect of TPM-bound LUKS on z15, by appennding the LUKS keys and `crypttab` to the `initramfs`, which is encrypted as part of `sd-boot`. Original `initramfs` under `/boot` is non-modified
```
bgilbert: `appending`

os/20211014-s390x-secure-execution.md
```diff
@@ -0,0 +1,190 @@
...
+ - generates `/se/sd-boot` encrypted by universal public hostkey (valid for all Z15 and newer machines)
+ - `zipl`es `/se/sd-boot`, so `qemu-secex.qcow2` image could only be run with Secure Execution
+
+Customer should add his public hostkeys(s) into `ignition` config.
+> **Note:** We cannot use LUKS `integrity` option, because it doesn's support resizing of partitions, so `rootfs` cannot be growned during `firstboot`.
```
bgilbert:
```suggestion
> **Note:** In this step we will not use the LUKS `integrity` option, because it doesn't support resizing of partitions, so `rootfs` cannot be grown during firstboot.
```

os/20211014-s390x-secure-execution.md
```diff
@@ -0,0 +1,190 @@
...
+
+## Step 3: During firstboot `/boot` and `rootfs` are reencrypted with randomly generated LUKS keys and `integrity` option, which is deffered from Step 2
```
bgilbert:
```suggestion
## Step 3: During firstboot `/boot` and `rootfs` are reencrypted with randomly generated LUKS keys; `integrity` option is enabled
```

os/20211014-s390x-secure-execution.md
```diff
@@ -0,0 +1,190 @@
...
+
+During firstboot the LUKS keys will be changed to a randomly generated ones. The idea behind this is to set on each host the LUKS keys to a unique keys during firstboot. This should prevent reading the keys from the provided pre-build image.
```
bgilbert: Add something like: "Also, LUKS integrity protection will be enabled both in the pre-built image and after re-encryption."

os/20211014-s390x-secure-execution.md
```diff
@@ -0,0 +1,190 @@
...
+
+## Step 4: Ignition Protection
+
+Ignition config will be built and when saved encrypted with a customer's public key.
```
bgilbert:
```suggestion
The user will encrypt the Ignition config with the Ignition public key for the pre-built image.
```

os/20211014-s390x-secure-execution.md
```diff
@@ -0,0 +1,190 @@
...
+The private key is part of the `sd-boot` image and is generated by RH/IBM during qcow2 image build. The public key is part of the build metadata.
+During firstboot the Ignition config gets decrypted and the key gets removed immediately after it. This could be done with a
```
bgilbert:
```suggestion
During firstboot the Ignition config gets decrypted and the key gets removed immediately after it.
```

os/20211014-s390x-secure-execution.md
nikita-dubrovskii: yes, during LUKS keys modification we'd have to do this

os/20211014-s390x-secure-execution.md
```diff
@@ -0,0 +1,241 @@
...
+The private key is part of the `sdboot` image and is randomly generated during qcow2 image build. The public key is part of the build metadata. During firstboot the Ignition config gets decrypted.
+We should avoid printing the Ignition config to the console (e.g. serial and VGA) and scrub it before any shell access is possible.
+
+- In the SE path, add dropins to disable Ignition logging
+- Right after config decryption drop the ignition's private key to avoid leaking it.
+- In the emergency shell code, delete `/run/ignition.json` before entering the shell.
+- User generates the symmetric `key` and encrypts with it the `config.ign`:
+  ```
+  openssl rand -base64 96 > key
+  openssl enc -salt -pbkdf2 -aes256 -in config.ign -out config.crypted -pass file:./key
```
bgilbert: It's awkward and error-prone to manually perform low-level operations such as generating the symmetric key.  We should provide tooling for this, or preferably use existing tools.  Is GPG a reasonable alternative?  The key management would be sort of awkward, though; the user would end up with a number of public keys for random OS images on their keyring.

os/20211014-s390x-secure-execution.md
```diff
@@ -0,0 +1,241 @@
...
+  ```
+- User encryptes his `key` (using `ignition.pub`) and generates a tarball:
+  ```
+  openssl pkeyutl -encrypt -in key -out key.crypted -pubin -inkey /path/to/ignition.pub
+  find . -mindepth 1 -name "*.crypted" | tar -cvf /path/to/ignition_crypted -T -
```
bgilbert: Doesn't this imply that the initramfs is unpacking an untrusted tar file?  That puts `tar` in a security-sensitive role, which doesn't seem ideal.  I'd strongly favor using existing hardened tools rather than assembling our own.

nikita-dubrovskii: tar here is only for convenience. we can add one more `virtio-blk` dev for a symmetric key

os/20211014-s390x-secure-execution.md
nikita-dubrovskii: ok, GPG may make life easier here. i'll update this and WIP-PRs

jlebon (approved):
So, we implemented this proposal now. Unless anyone is opposed, I think we should merge this PR since it documents the work that was done. I see two comment threads that are still not marked as resolved. @nikita-dubrovskii Are there outstanding items there that we should track somewhere or are those resolved now?

nikita-dubrovskii:
> Are there outstanding items there that we should track somewhere or are those resolved now?

No, marked all as resolved.

PR #8, opened 2021-11-22
coreos-layering: Three minor tweaks

cgwalters:
- xref new Fedora Change
- Fix base image name
- More info on nesting container builds


jlebon (approved):
While we're here, let's also rename the file? I noticed other enhancements in this repo have a timestamp in their names.

cgwalters:
> While we're here, let's also rename the file? I noticed other enhancements in this repo have a timestamp in their names.

Hmm, what is the rationale for timestamp in the filename actually?  It looks like the Rust equivalent uses a serial number, and there are no serial or timestamps in openshift/enhancements.

cgwalters:
Merging as is because I have already added the existing URL in a few places, and I don't quite want to go through and update all those links until there's a rationale for it.

PR #10, opened 2023-10-24
s390x: zVM Secure IPL enablement

nikita-dubrovskii:


os/20231024-s390x-zvm-secure-boot.md
```diff
@@ -0,0 +1,75 @@
...
+- SCSI devices.
+- ECKD devices.
+
+---
+
+## Prerequisites
+
+### Security keys are loaded onto the HMC certificate store
+
+It is the customer's responsibility to ensure that the security keys that are required by a guest are present on the destination system.Security keys are required on all members of a relocation domain where a guest might be relocated and later IPLed with the secure option. For more information, see [How to enable secure boot on the HMC](https://www.ibm.com/support/pages/how-enable-secure-boot-hmc).
```
jlebon:
```suggestion
It is the customer's responsibility to ensure that the security keys that are required by a guest are present on the destination system. Security keys are required on all members of a relocation domain where a guest might be relocated and later IPLed with the secure option. For more information, see [How to enable secure boot on the HMC](https://www.ibm.com/support/pages/how-enable-secure-boot-hmc). 
```

os/20231024-s390x-zvm-secure-boot.md
```diff
@@ -0,0 +1,75 @@
+# IBM Secure Boot for Linux
```
jlebon: Minor: this might mislead one to believe that this is Secure Boot but really it's more IBM's version of the same idea, right? I wonder if it'd be clearer to rename this e.g.

```suggestion
# z/VM Guest Secure IPL (Secure Boot-like protection)
```

?

os/20231024-s390x-zvm-secure-boot.md
```diff
@@ -0,0 +1,75 @@
...
+
+z/VM® supports guest secure IPL. Guest secure IPL supports the NIAP (National Information Assurance Partnership) operating system protection profile, which supports the Common Criteria certification.
+
+A z/VM user can request that the machine loader validate the signed IPL code by using the security keys that were previously loaded by the customer into the HMC certificate store. The validation ensures that the IPL code is intact, unaltered, and originates from a trusted build-time source.
+
+This support provides the ability for a Linux guest to exploit hardware to validate the code being booted, helping to ensure it is signed by the client or its supplier.
+
+Support is provided for the following device types:
+- SCSI devices.
+- ECKD devices.
```
jlebon: This looks like a copy/paste from somewhere? Which is fine, but would be good to include the link to it.

os/20231024-s390x-zvm-secure-boot.md
```diff
@@ -0,0 +1,75 @@
...
+
+---
+
+## Prerequisites
+
+### Security keys are loaded onto the HMC certificate store
+
+It is the customer's responsibility to ensure that the security keys that are required by a guest are present on the destination system.Security keys are required on all members of a relocation domain where a guest might be relocated and later IPLed with the secure option. For more information, see [How to enable secure boot on the HMC](https://www.ibm.com/support/pages/how-enable-secure-boot-hmc). 
+
+### The IPL device number and the SECURE option load parameters are set
```
jlebon: This seems to detail how to turn on Guest Secure IPL day 2, which is much less interesting than day 1. I almost think we should remove that section entirely.

The bit about verifying that RHCOS booted in that mode is helpful though. We could keep that in a separate section under the Proposal header.

os/20231024-s390x-zvm-secure-boot.md
```diff
@@ -0,0 +1,75 @@
...
+   $ dmesg | grep "Secure-IPL enabled"
+   [    0.029829] setup: Linux is running with Secure-IPL enabled
+   ```
+
+# Proposal: Enabling Secure Boot at installation time
+
+Assuming zVM is ready for secure boot, we can setup LOADDEV at installation time
+
+## coreos-installer
+1) Add new `coreos.inst.secureboot` karg and `--with-secure-boot` option. `coreos-installer-generator` appends the switch when karg is provided
```
jlebon: Similar concern here, I think calling it `coreos.inst.secureboot` might lead to confusion and users thinking this applies to other arches with actual Secure Boot.

Maybe `coreos.inst.secureipl` and `--with-secure-ipl`?

os/20231024-s390x-zvm-secure-boot.md
```diff
@@ -0,0 +1,75 @@
...
+   ```
+   $ dmesg | grep "Secure-IPL enabled"
+   [    0.029829] setup: Linux is running with Secure-IPL enabled
+   ```
+
+# Proposal: Enabling Secure Boot at installation time
+
+Assuming zVM is ready for secure boot, we can setup LOADDEV at installation time
+
+## coreos-installer
```
jlebon: Minor: space

os/20231024-s390x-zvm-secure-boot.md
```diff
@@ -0,0 +1,75 @@
...
+1) Add new `coreos.inst.secureboot` karg and `--with-secure-boot` option. `coreos-installer-generator` appends the switch when karg is provided
+2) During isntallation we check for `--with-secure-boot` and use `vmcp` tool to set LOADDEV
```
jlebon:
```suggestion
2) During installation we check for `--with-secure-boot` and use `vmcp` tool to set LOADDEV
```

os/20231024-s390x-zvm-secure-boot.md
```diff
@@ -0,0 +1,75 @@
...
+3) Add new systemd unit `coreos-installer-reboot-secure.service` to restart from LOADDEV or modify `coreos-installer-reboot.service` to do the same (which immediatly termintes running CoreOS VM)
```
jlebon: I.e. a simple `systemctl reboot` won't work? It needs to be done via `cp ipl loaddev`? Is this true on reboots of the installed system too?

nikita-dubrovskii: Unfortunately we must use `cp ipl loaddev`. Luckily any subsequent `reboot` just works (until we shutdown zVM itself)

os/20231024-s390x-zvm-secure-boot.md
nikita-dubrovskii: where? ))

os/20231024-s390x-zvm-secure-ipl.md
```diff
@@ -0,0 +1,72 @@
...
+
+# Enabling Secure IPL at installation time (day-1)
+
+Assuming zVM is ready for secure boot, we can setup LOADDEV at installation time
+
+## coreos-installer
+
+1) Add new `coreos.inst.secure_ipl` karg and `--secure-ipl` option. `coreos-installer-generator` appends the switch when karg is provided
+2) During installation we check for `--secure-ipl` and use `vmcp` tool to set LOADDEV
+3) Add new systemd unit `coreos-installer-reboot-loaddev.service` to restart from LOADDEV (which immediatly termintes running CoreOS VM)
```
jlebon:
```suggestion
3) Add new systemd unit `coreos-installer-reboot-loaddev.service` to restart from LOADDEV (which immediately terminates running CoreOS VM)
```

os/20231024-s390x-zvm-secure-ipl.md
```diff
@@ -0,0 +1,72 @@
...
+
+    Assuming RHCOS is installed on FCP disk `0.0.8007,0x500507630400d1e3,0x4001404c00000000`, from zVM execute:
+    ```
+    # cp set loaddev dev 8007 portname 50050763 0400d1e3 lun 4001404c 00000000 secure
+    ```
+3) Update bootloader
+    ```
+    $ sudo mount -o remount,rw /boot
+    $ sudo zipl -V --secure 1
+    $ sudo mount -o remount,ro /boot
```
jlebon:
```suggestion
    $ sudo unshare --mount bash -c 'mount -o remount,rw /boot && zipl -V --secure 1'
```

Avoids remounting rw for everyone and having to unmount.

os/20231024-s390x-zvm-secure-ipl.md
```diff
@@ -0,0 +1,72 @@
...
+
+Assuming zVM is ready for secure boot, we can setup LOADDEV at installation time
+
+## coreos-installer
+
+1) Add new `coreos.inst.secure_ipl` karg and `--secure-ipl` option. `coreos-installer-generator` appends the switch when karg is provided
+2) During installation we check for `--secure-ipl` and use `vmcp` tool to set LOADDEV
+3) Add new systemd unit `coreos-installer-reboot-loaddev.service` to restart from LOADDEV (which immediatly termintes running CoreOS VM)
+
+# Enabling Secure IPL on existing system (day-2)
```
jlebon, line 28: Maybe here something like:

> This is for informational purposes only. We expect users to only turn on Guest Secure IPL at installation time.

Right?

nikita-dubrovskii: Yes, added this section just for info

jlebon (approved):

